PR #1, opened 2015-06-04
Added a GetRow function, and modified hrpc.Get to hold families.

dgonyeo:
I've added the function GetRow to client.go, which will return a row given a table, key, and a set of families and qualifiers.

I also modified hrpc.Get to store a mapping of families to lists of qualifiers, which is used when serializing the request.

### DIFF
--- a/client.go
+++ b/client.go
@@ -32,7 +32,9 @@ var (
 		StopKey:    []byte{},
 	}
 
-	infoFamily = []byte("info")
+	infoFamily = map[string][]string{
+		"info": nil,
+	}
 )
 
 // region -> client cache.
@@ -111,7 +113,16 @@ func NewClient(zkquorum string) *Client {
 
 // CheckTable returns an error if the given table name doesn't exist.
 func (c *Client) CheckTable(table string) (*pb.GetResponse, error) {
-	resp, err := c.sendRpcToRegion(hrpc.NewGetStr(table, "theKey"))
+	resp, err := c.sendRpcToRegion(hrpc.NewGetStr(table, "theKey", nil))
+	if err != nil {
+		return nil, err
+	}
+	return resp.(*pb.GetResponse), err
+}
+
+// GetRow returns a single row fetched from hbase
+func (c *Client) GetRow(table string, rowkey string, families map[string][]string) (*pb.GetResponse, error) {
+	resp, err := c.sendRpcToRegion(hrpc.NewGetStr(table, rowkey, families))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


hrpc/get.go
```diff
@@ -56,13 +56,16 @@ func (g *Get) Serialize() ([]byte, error) {
...
 		},
 	}
-	if g.family != nil {
-		get.Get.Column = []*pb.Column{
-			&pb.Column{
-				Family: g.family,
-				//Qualifier: [][]byte{[]byte("theCol")},
-			},
+	for family, qualifiers := range g.families {
+		bytequals := make([][]byte, len(qualifiers), len(qualifiers))
```
tsuna: This is equivalent to saying just `bytequals := make([][]byte, len(qualifiers))` – by default length==capacity for new slices.

dgonyeo: My bad, I was unaware. Should be fixed now.

client.go
```diff
@@ -111,7 +113,16 @@ func NewClient(zkquorum string) *Client {
...
 func (c *Client) CheckTable(table string) (*pb.GetResponse, error) {
-	resp, err := c.sendRpcToRegion(hrpc.NewGetStr(table, "theKey"))
+	resp, err := c.sendRpcToRegion(hrpc.NewGetStr(table, "theKey", nil))
+	if err != nil {
+		return nil, err
+	}
+	return resp.(*pb.GetResponse), err
+}
+
+// GetRow returns a single row fetched from hbase
```
tsuna: Nit: `s/hbase/HBase/`

client.go
```diff
@@ -111,7 +113,16 @@ func NewClient(zkquorum string) *Client {
...
+func (c *Client) GetRow(table string, rowkey string, families map[string][]string) (*pb.GetResponse, error) {
```
tsuna: Not for later: we should have an API that accepts mostly `[]byte`, but it's also nice to provide an additional function that accepts `string` arguments, because it's a common thing to use too.

tsuna:
Merged as tsuna/gohbase@8c9bddd27a6c60227c0a4c8e52e2c6a4b238c3c7, thanks!

client.go
dgonyeo: My current solution relies on using those as keys to a map, and Go does not allow using slices as keys in maps. I'll figure out something else tomorrow so we can accept tables and rowkeys as []byte.

tsuna: Ah, yeah, that's annoying.  We can also chose to store things as strings internally... A string is mostly an immutable `[]byte` in Go (?) – but we'll have to keep copying things back and forth..?

PR #2, opened 2015-06-08
Mutations and beginnings of scan

dgonyeo:
I'll keep PRs smaller after this one.

This includes hrpc/mutate.go, which includes the RPC calls for some of the things supported by the Mutate protobuf message. Most notably there are now calls for Put, Delete, Append, and Increment. Increment is not currently working (hbase returns some error when it is called).

This also includes some of the initial code for Scan. It fetches the first set of records for the scan from hbase and then stops (without closing the scanner).

### DIFF
--- a/client.go
+++ b/client.go
@@ -129,15 +129,15 @@ func (c *Client) Get(table string, rowkey string, families map[string][]string) 
 	return resp.(*pb.GetResponse), err
 }
 
-/*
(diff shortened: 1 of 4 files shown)


tsuna:
Pushed amended versions as 0c7e48c6a35579b481f4de0370ea28322f662031 and 076d18a06be2c20dde27d009d3c090466a8ebddc, thanks!

PR #3, opened 2015-06-08
Multithread writes

dgonyeo:
This PR changes region/client.go to have a separate writer thread for each client. When a thread wishes to write, it writes its buffer to an unbuffered channel that the writer thread is reading from. If the writer encounters an error, it closes the done channel, which all threads that are writing to the send channel will be watching. Upon the done channel being closed, threads consult sendErr for the reason. The sendWrite function implements the intended behavior for functions wishing to write a buffer.

### DIFF
--- a/client.go
+++ b/client.go
@@ -129,15 +129,15 @@ func (c *Client) Get(table string, rowkey string, families map[string][]string) 
 	return resp.(*pb.GetResponse), err
 }
 
-/*
-func (c *Client) NewScan(table string, families map[string][]string, startRow, stopRow *[]byte) ([]*pb.ScanResponse, error) {
+// Scan will retrieve the values specified in families from multiple rows in
+// the given hbase table.
+func (c *Client) Scan(table string, families map[string][]string, startRow, stopRow []byte) (*pb.ScanResponse, error) {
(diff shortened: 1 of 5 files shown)


tsuna:
BTW, `make check` is failing for me because of `golint` complaints.  I don't know why Travis CI sees `make` exit 0 despite golint printing the same complaints there.

```
region/client.go:19:2: error var ShortWriteErr should have name of the form ErrFoo
region/client.go:19:2: exported var ShortWriteErr should have comment or be unexported
make: *** [lint] Error 1
```

If anyone has any idea of what's wrong with how Travis CI executes the tests, please let me know.

tsuna:
I rejiggered your commit and pushed them here:

https://review.gerrithub.io/236108 Factor out some duplicated code.
https://review.gerrithub.io/236109 Put a mutex around write calls.

Can you please take a look?  If you're OK, I'll merge them from there and we'll close this PR.

dgonyeo:
LGTM

tsuna:
Merged as 071b4fbc492254c43bc6ee57f2d620a33a585771 and 1a85cb715278d7b28da50e83b7ef5d326697eae1, thanks!

PR #19, opened 2016-11-07
Key of regions cache is the name, not start key

cannium:


Timoha:
fixed in https://github.com/tsuna/gohbase/commit/29d1d78bc0eb58da9eb390cac2b898a0642661e5. Closing the Pull request.

PR #21, opened 2016-12-05
Added support for alternative zookeeper path roots.

jonbonazza:


ipostelnik:
@Timoha can you merge this PR? Required for people running more than one cluster with single zookeeper.

zk/client.go
```diff
@@ -33,16 +33,21 @@ func init() {
 // in ZooKeeper
 type ResourceName string
 
+// Prepend creates a new ResourceName with prefix prepended to the former ResourceName.
+func (r ResourceName) Prepend(prefix string) ResourceName {
+	return ResourceName(prefix + string(r))
```
Timoha: should use [`filepath.Join`](https://golang.org/pkg/path/filepath/#Join) in order to avoid double slashes if `ZookeeperRoot` has a trailing slash.

jonbonazza:
Unfortunately, I already deleted this fork. I will reproduce the changes now and open a separate PR.

PR #26, opened 2017-02-03
Fix panic on removing last entry in clientRegionCache

ipostelnik:


Timoha:
Can probably close this since changed to use a map to hold regions: https://github.com/tsuna/gohbase/commit/e0df983025e2101c6af7c98e3efef6a4dbd1483c

PR #31, opened 2017-02-15
Made scan work across regions, respect numberOfRows limit

### LINKED ISSUE #27: Scan doesn't work across regions, doesn't respect numberOfRows filter
We've fixed it on our fork here - https://github.com/cloudflare/gohbase/pull/8. What is the current way to submit PRs?

ipostelnik:
Fixes #27

### DIFF
--- a/client.go
+++ b/client.go
@@ -142,19 +142,20 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
 	maxVersions := s.MaxVersions()
 	numberOfRows := s.NumberOfRows()
 	for {
-		// Make a new Scan RPC for this region
 		if rpc != nil {
 			// If it's not the first region, we want to start at whatever the
 			// last region's StopKey was
 			startRow = rpc.RegionStop()
 		}
+		regionRowLimit := numberOfRows - uint32(len(results))
 
 		// TODO: would be nicer to clone it in some way
-		rpc, err := hrpc.NewScanRange(ctx, table, startRow, stopRow,
+		var err error
+		rpc, err = hrpc.NewScanRange(ctx, table, startRow, stopRow,
 			hrpc.Families(families), hrpc.Filters(filters),
 			hrpc.TimeRangeUint64(fromTs, toTs),
 			hrpc.MaxVersions(maxVersions),
-			hrpc.NumberOfRows(numberOfRows))
+			hrpc.NumberOfRows(regionRowLimit))
 		if err != nil {
 			return nil, err
 		}
@@ -163,6 +164,7 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
 		if err != nil {
 			return nil, err
 		}
+
 		scanres = res.(*pb.ScanResponse)
 		results = append(results, scanres.Results...)
 
@@ -170,7 +172,7 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
 		// true, so we should figure out if there's a better way to know when
 		// to move on to the next region than making an extra request and
 		// seeing if there were no results
-		for len(scanres.Results) != 0 {
+		for len(scanres.Results) != 0 && uint32(len(results)) < numberOfRows {
 			rpc = hrpc.NewScanFromID(ctx, table, *scanres.ScannerId, rpc.Key())
 
 			res, err = c.sendRPC(rpc)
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -163,14 +163,16 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
...
 		scanres = res.(*pb.ScanResponse)
 		results = append(results, scanres.Results...)
 
 		// TODO: The more_results field of the ScanResponse object was always
 		// true, so we should figure out if there's a better way to know when
 		// to move on to the next region than making an extra request and
 		// seeing if there were no results
-		for len(scanres.Results) != 0 {
+		for len(scanres.Results) != 0 &&
+				(numberOfRows == 0 || uint32(len(results)) < numberOfRows) {
```
Timoha: `numberOfRows == 0` should probably be checked before even sending the first request. Also, the case of "if there are results and numberOfRows == 0" doesn't really make sense, so maybe just drop it?

client.go
```diff
@@ -188,13 +190,16 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
...
-		// greater than or equal to the stop_key of this scanner provided
-		// that (2) we're not trying to scan until the end of the table).
-		// (1)
-		if len(rpc.RegionStop()) == 0 ||
+		// because:
+		// (0) we got the number of rows user asked for
+		// (1) it's the last region or
+		// (3) because its stop_key is greater than or equal to the stop_key of this scanner,
+		// provided that (2) we're not trying to scan until the end of the table).
+		if uint32(len(results)) >= numberOfRows ||
```
Timoha: if this "if" statement is false, and we are to move to next region, `numberOfRows` should probably be updated to `numberOfRow - uint32(len(results))`?

client.go
ipostelnik: `numberOfRows == 0` means there's no limit

client.go
ipostelnik: yes, good point

client.go
Timoha: > `numberOfRows == 0` means there's no limit

could you please provide a reference to this? I'm looking at asynchbase (since gohbase is modeled from it) and it looks like they don't even allow a value of 0: https://github.com/OpenTSDB/asynchbase/blob/master/src/Scanner.java#L460

ipostelnik: I was just going by current implementation. However, since it's a unit32, `Scan.numberOfRows` will have default value of 0 if omitted. Maybe `Scan.Serialize()` should omit setting it when 0?

Timoha: the default value `DefaultNumberOfRows = 128` is set here: https://github.com/tsuna/gohbase/blob/master/hrpc/scan.go#L64
I'm thinking the right way would be to check if the value is strictly positive when it passed as an option here: https://github.com/tsuna/gohbase/blob/master/hrpc/call.go#L211

Sorry that these values are all over the place in gohbase/hrpc.

client.go
```diff
@@ -163,14 +164,15 @@ func (c *client) Scan(s *hrpc.Scan) ([]*hrpc.Result, error) {
...
+
 		scanres = res.(*pb.ScanResponse)
 		results = append(results, scanres.Results...)
 
 		// TODO: The more_results field of the ScanResponse object was always
 		// true, so we should figure out if there's a better way to know when
 		// to move on to the next region than making an extra request and
 		// seeing if there were no results
-		for len(scanres.Results) != 0 {
+		for len(scanres.Results) != 0 && uint32(len(results)) < numberOfRows {
```
Timoha: My bad, I'm realizing now that` numberOfRows` option is meant to optimize how many rows at max should be returned in each ScanResponse (for each RPC), and it's not meant to limit number of rows returned for the global Scan: http://opentsdb.github.io/asynchbase/javadoc/index.html. Otherwise, when we do scan between 2 keys, we always will have to know how many rows there are between them. What do you think?

hrpc/call.go
```diff
@@ -208,12 +208,16 @@ func MaxVersions(versions uint32) func(Call) error {
 
 // NumberOfRows is used as a parameter for request creation.
 // Adds NumberOfRows constraint to a request.
```
Timoha: which means this comment is wrong

client.go
ipostelnik: I don't think this is a problem. Before the changes, scanning between 2 keys should have returned ~128 rows. I interpreted the parameters as `select ... where key between startKey and stopKey limit numberOfRows`.

ipostelnik: Overall the Client.Scan API should be redesigned to use a channel or, at least, something that looks like an iterator over slices  of rows.

Timoha: I agree that it should be redesigned to be able to control when to stop scanning. Something like [Scanner.nextRows()](http://opentsdb.github.io/asynchbase/javadoc/index.html) as in asynchbase should do it. 

At the moment, we are going to be limited to 128 rows in a scan, which seems incorrect.
However, If we are to remove the `uint32(len(results)) < numberOfRows` check, and someone decided to retrieve the whole key space of a table, the gohbase will do that and might run out of memory, which is also incorrect.

I'm ok with merging your fix, and then just working on redesigning the scan Api. Lets see what @tsuna thinks.

ipostelnik: This PR did not change existing behavior of returning 128 rows by default. Given existing bugs I suspect I'm the only user of the scan feature on any meaningful data.

Timoha: alright, looks like @tsuna is occupied, I'll merge it, thanks!

PR #34, opened 2017-02-22
Fixed issue where admin client didn't have its default zookeeper path set.

jonbonazza:


jonbonazza:
@Timoha  here is the fix you requested. Looks like it was only missing in the admin client.

jonbonazza:
Also, I confirmed it builds and all unit tests pass.

Timoha:
we also got integration tests (https://github.com/tsuna/gohbase/blob/master/integration_test.go) which actually caught the issue on our end. We'll have to find some way to run those before merging pull requests from now on. If you have hbase installed on your machine, you can run them with `make integration`

PR #36, opened 2017-02-22
Fix effective user for accessing region servers.

jonbonazza:
@Timoha here is another one of the minor fixes I mentioend.

### DIFF
--- a/region/client.go
+++ b/region/client.go
@@ -101,6 +101,8 @@ type client struct {
 
 	rpcQueueSize  int
 	flushInterval time.Duration
+
+	effectiveUser string
 }
 
 type call struct {
@@ -359,7 +361,7 @@ func (c *client) readFully(buf []byte) error {
 func (c *client) sendHello(ctype ClientType) error {
 	connHeader := &pb.ConnectionHeader{
 		UserInfo: &pb.UserInformation{
-			EffectiveUser: proto.String("gopher"),
+			EffectiveUser: proto.String(c.effectiveUser),
 		},
 		ServiceName: proto.String(string(ctype)),
 		//CellBlockCodecClass: "org.apache.hadoop.hbase.codec.KeyValueCodec",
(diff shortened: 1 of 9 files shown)


region/client.go
```diff
@@ -359,7 +359,7 @@ func (c *client) readFully(buf []byte) error {
 func (c *client) sendHello(ctype ClientType) error {
 	connHeader := &pb.ConnectionHeader{
 		UserInfo: &pb.UserInformation{
-			EffectiveUser: proto.String("gopher"),
+			EffectiveUser: proto.String("root"),
```
Timoha: should we have an option to be able to specify this when we create a client instead?

jonbonazza: An option is probably better in the long run. I will add that in a bit instead. This was mostly a quick fix as root is a more likely user than gopher ;).

Thinking about it though, it should be trivial to make this configurable with options. Let me redo this one.

jonbonazza:
@Timoha Did I miss anything here?

Timoha:
I think you forgotten to update `admin_client.go` as well. Also, take a look at the failed build in Travis, we have small check to see if lines are longer then 100 chars, looks like you have some.

jonbonazza:
@Timoha Ahhh the admin_client.. my nemesis :P Anyway, I updated it to set the new field to the default. Also, I broke up the lines that the linter was complaining about. 

With all that said, does travisci not rerun on a force pushed amended commit? Still seems to be using the old code in the latest travisci build. Anyway to force a run?

Timoha:
> With all that said, does travisci not rerun on a force pushed amended commit? Still seems to be using the old code in the latest travisci build. Anyway to force a run?

Clicking on "Details" of your Travis build it looks like you have a different error in tests now, so it did get triggered

jonbonazza:
@Timoha fixed the unit test issue. All green now. =) I misinterpreted the CI logs before. Apologies.

Timoha:
@jonbonazza could you please rebase this against the master, I added integration tests, so would like all PRs go through them

jonbonazza:
@Timoha I rebased as you requested. Looking at the test details I see some ERR lines in the logs, but TravisCI says the tests passed, so I guess those logs are expected?

Timoha:
yes, it ERR should be fine, when we close hbase client in the end of the test, it closes all the region clients. And those are logging with ERR as if they failed, it's all good.

PR #50, opened 2017-06-19
adding cluster status request to admin client

barnjamin:


barnjamin:
Signed the contrib agreement, added a sanity check function.  Is this a reasonable pattern to follow for the rest of the admin like functions?  I intend to add at least a couple more like assigning/compacting regions.

integration_test.go
```diff
@@ -29,6 +29,23 @@ var host = flag.String("host", "localhost", "The location where HBase is running
 
 const table = "test1"
 
+// ClusterStatus gets a summary of relevant cluster stats
+func ClusterStats(client gohbase.AdminClient) error {
```
Timoha: Could you please make it a full test, something like this:
```
func TestClusterStatus(t *testing.T) {
      ac := gohbase.NewAdminClient(*host)
      defer ac.Close()
      stats, err := ac.ClusterStatus()
      if err != nil {
          t.Fatal(err)
      }
      if len(stats.GetMaster().GetHostName()) == 0 {
           t.Fatal("Master hostname is empty in ClusterStatus")
      }
}
```

Timoha:
Yes, this looks great, add whatever you need as it's likely that others would need it as well.
Note that travis builds are failing for your change (linting like forgot to add comments to exported functions).

barnjamin:
Sorry for the spam

barnjamin:
Closing for #53

PR #52, opened 2017-06-21
Added exposing storeLimit and storeOffset for gets and scans

dougklein:
This is very useful if you make active use of column qualifiers and have many sorted columns columns per key


hrpc/call.go
```diff
@@ -205,6 +205,49 @@ func MaxVersions(versions uint32) func(Call) error {
 	}
 }
 
+// MaxResultsPerColumnFamily sets the maximum number of values returned for each row
```
Timoha: would it be more precise to say: "MaxResultsPerColumnFamily is an option for Get or Scan requests that sets the maximum number of cells returned per each column family in a row"?

hrpc/call.go
```diff
@@ -205,6 +205,49 @@ func MaxVersions(versions uint32) func(Call) error {
...
+			if maxresults > math.MaxInt32 {
+				return errors.New(opt + "exceeds supported number of value results")
+			}
+			c.storeLimit = maxresults
+		}
+		return nil
+	}
+}
+
+// ResultOffset sets the offset within a column family set of results
```
Timoha: how about "ResultOffset is a option for Scan or Get requests that sets the offset for cells within a column family"

hrpc/scan.go
```diff
@@ -29,6 +29,8 @@ const (
 	DefaultMaxResultSize = 2097152
 	// DefaultNumberOfRows is default maximum number of rows fetched by scanner
 	DefaultNumberOfRows = math.MaxInt32
+	// DefaultMaxResultsPerColumnFamily is the default max number of columns fetched for each row
```
Timoha: "DefaultMaxResultsPerColumnFamily is the default max number of cells fetched per column family for each row "

dougklein:
Your suggestions are great - I was a bit sloppy on the comments. Will put those in.

I signed the contributor agreement and hit submit; I'm not sure where it went but am hoping that all went where it needed to.

I'm trying to get past the codecov audits; it's claming the tests aren't covering the code, yet they are being called from integration_test.go. It's just they are called quite far down the call stack. I'm going to try to call those outside of the normal context and see how I fare. If you have suggestions to help this along I would appreciate it. 

Cheers!

Timoha:
Thanks!
I wouldn't worry about code coverage checks failing, as you are right, the integration test don't show up in there (that's why you only see 41% coverage on README, while those actually only unit tests within each specific package).

Right now the only issue you have is merge conflicts with master, once you resolve those, I merge your changes.

dougklein:
I think I have the merge conflicts resolved. If not can you let me know?

Timoha:
Nope, I'm still getting `This branch cannot be rebased due to conflicts`. Your commit looks empty btw: https://github.com/tsuna/gohbase/pull/52/commits/7e0b6f5f79ff68d0ba9735a3447c548fbc8010a8

dougklein:
Sorry about this; I've not done this before. I think I've just managed to correctly rebase my fork against master. If this doesn't work I may just create a new fork and re-apply the changes.

Timoha:
have you tried something like `git pull --rebase https://github.com/tsuna/gohbase.git master`? basically you don't want to have merge commits, instead you want your changes to be applied on top of the current master modifying them if there are conflicts. You should look into https://git-scm.com/book/en/v2/Git-Branching-Rebasing.

dougklein:
I thought I had done that. In the end I deleted my fork and re-created it, applying my changes over the top of the latest master.

PR #53, opened 2017-06-22
Branch 1.3

barnjamin:
Adding proto files from branch-1.3


Timoha:
Merged, thanks for contributing. A suggestion for next time is try to separate different changes to different PRs. For example, updating proto files is one PR and adding ClusterStatus is another one, thanks!

PR #55, opened 2017-06-27
Compact region

barnjamin:
Something like what is in the tools directory is what I'm trying to run and getting UnsupportedOp


hrpc/compact.go
```diff
@@ -0,0 +1,57 @@
...
+type CompactRegion struct {
+	tableOp
+
+	region string
+	major  bool
+	family []byte
+}
+
+//Create a new CompactRegion struct
+func NewCompactRegion(region string, family []byte, major bool) *CompactRegion {
```
Timoha: you should provide the context as the first argument.
also, is there a reason for "region" being a string and "family" being []byte?

hrpc/compact.go
```diff
@@ -0,0 +1,57 @@
...
+
+type CompactRegion struct {
+	tableOp
+
+	region string
+	major  bool
+	family []byte
+}
+
+//Create a new CompactRegion struct
```
Timoha: the comment should be of format `NewCompactRegion ..`

hrpc/compact.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		region: region,
+		family: family,
+		major:  major,
+	}
+}
+
+//Return the name of this RPC  method
+func (cr *CompactRegion) Name() string {
+	return "CompactRegion"
```
Timoha: my suspicion is this "CompactRegion" is a wrong name for the rpc, because that's one of the places where `UnsupportedOperationException` is thrown: https://github.com/apache/hbase/blob/branch-1.3/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcServer.java#L1974

hrpc/compact.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+//Create a new CompactRegion struct
+func NewCompactRegion(region string, family []byte, major bool) *CompactRegion {
+	chunks := strings.Split(region, ",")
+	return &CompactRegion{
+		tableOp: tableOp{
+			base: base{
+				ctx:   context.Background(),
+				table: []byte(chunks[0]),
```
Timoha: should verify that you have 2 chunks, otherwise if there's no "," it will panic

hrpc/compact.go
barnjamin: I believe it should match the name of the method in the service section of the proto files: https://github.com/apache/hbase/blob/branch-1.3/hbase-protocol/src/main/protobuf/Admin.proto#L284

barnjamin:
Sorry for wasting your time reviewing this, I just threw this together for an example.

PR #57, opened 2017-06-28
Add support for MaxResultsPerColumnFamily and ResultOffset

dougklein:
Re-based off main branch and added changes

### DIFF
--- a/hrpc/call.go
+++ b/hrpc/call.go
@@ -205,6 +205,51 @@ func MaxVersions(versions uint32) func(Call) error {
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Timoha:
Looks great, thank you for the contribution and the tests!

timoha:
I just realized that this was probably possible with use of filters? https://godoc.org/github.com/tsuna/gohbase/filter#ColumnPaginationFilter
they are all undocumented, but just by reading the signatures it feels like they could be used as well, might be pretty powerful.

dougklein:
Looking at it, I think I might have been able to use this. Oh well. :-)

PR #60, opened 2017-07-12
Added Scanner.ScanCh for creating a channel to consume rows as they a re scanned.

jonbonazza:
ScanCh is a convenience method for retrieving a buffered channel that can be used to consume rows as they are scanned. As the comment mentions, errors are swallowed and this should only be used when you don't really care about errors.

As a potential enhancement, ```ScanCh``` could take a variadic argument of callback functions for when an error is found. This would allow us to gain the benefit of error handling as well as the convenience of a channel. This is an idiom that I didn't see elsewhere in the codebase, however, so I wanted to check with everyone before implementing it.

### DIFF
--- a/hrpc/scan.go
+++ b/hrpc/scan.go
@@ -37,6 +37,10 @@ const (
 // Scanner will be automatically closed if there's no more data to read,
 // otherwise Close method should be called.
 type Scanner interface {
+	// ScanCh returns a channel for consuming rows as they are scanned. If an error is
+	// encountered, that row is skipped and the error is ignored. This method should only
+	// be used if errors are not a concern.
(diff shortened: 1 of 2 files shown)


jonbonazza:
I'll write unit tests soon.

timoha:
This seems to be a simple wrapper around the Next() call, is there a reason this needs to be in the interface and not just be a helper function in your code? IMHO, it might confuse the users about which should be used and it doesn’t return an error. Any inputs @tsuna?

jonbonazza:
@timoha It can just as easily be a helper function in the code, and in fact, that's what we currently do, but I figured this would be useful to others so I included it here. If you don't agree, feel free to close the PR. =)

PR #61, opened 2017-07-18
Adding TTL for Mutate Requests

barnjamin:
Adding this here now to get your opinion.  If it looks good I'll add some tests.

I was surprised to see that the way this TTL works is that after the time has passed, the cell will be gone _or_ show the previous value if one was overwritten, even with versions set to 1.


hrpc/mutate.go
```diff
@@ -376,6 +409,18 @@ func (m *Mutate) serializeWithReflect() (*pb.MutateRequest, error) {
 	if m.timestamp != MaxTimestamp {
 		mProto.Timestamp = &m.timestamp
 	}
+
+	if m.ttl != nil && len(m.ttl) > 0 {
```
timoha: no need for `m.ttl != nil &&`, `len()` covers it

hrpc/mutate.go
```diff
@@ -376,6 +409,18 @@ func (m *Mutate) serializeWithReflect() (*pb.MutateRequest, error) {
...
+		if mProto.Attribute == nil {
+			mProto.Attribute = []*pb.NameBytesPair{}
```
timoha: no need for this also, if mProto.Attribute is nil, append will create a new slice.

hrpc/mutate.go
```diff
@@ -305,6 +326,18 @@ func (m *Mutate) serializeNoReflect() *pb.MutateRequest {
...
```
timoha: same here

PR #62, opened 2017-07-19
Add namespace support for table creation

mindon:
parse namespace from the table name separated by ":"


hrpc/create.go
```diff
@@ -91,12 +92,19 @@ func (ct *CreateTable) ToProto() (proto.Message, error) {
...
+	table := ct.table
+	i := bytes.Index(table, []byte(":"))
+	if i > -1 {
+		namespace = table[:i]
+		table = table[i+1:]
+	}
 	return &pb.CreateTableRequest{
 		TableSchema: &pb.TableSchema{
 			TableName: &pb.TableName{
 				// TODO: handle namespaces
```
timoha: the comment is not needed anymore

PR #66, opened 2017-08-02
Expose reversed option on scan

dougklein:
Useful scan functionality especially when a zero padded timestamp is part of a row key so that you can retrieve rows in reverse order.

### DIFF
--- a/hrpc/scan.go
+++ b/hrpc/scan.go
@@ -73,6 +73,7 @@ type Scan struct {
 
 	closeScanner        bool
 	allowPartialResults bool
+	reversed            bool
 }
 
 // baseScan returns a Scan struct with default values set.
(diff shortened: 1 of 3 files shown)


timoha:
I think this won't work over regions. The reason the test doesn't fail is probably because you only have one region. You might need to modify the next startrow to be the row before the current region's StartKey()?

dougklein:
You are correct; I pre-split a table and ran the test and it doesn't region change. To be honest, for our specific use case we are just trying to get the upper and lower bounds of known min and max ranges. I can use the NewGetBefore function for this. If I have time I'll look at re-opening this unless someone else feels like taking it over.

timoha:
I actually have reverse scanner on my TODO list as hbase maintainers are planning to make GetRowBefore obsolete. I’ll spend some time soon on it, thanks for starting this though!

dougklein:
Alas I will have to have a look at this as the GetRowBefore returns nothing on our HBase cluster (1.3.1). Although I only need the last (first) row, it is possible that there would be a region split between my start key and the first value in the range.

PR #67, opened 2017-08-09
Variable/Optional Timeouts for Zookeeper, Region Lookups, and Region Reads

travishegner:
In our environment we are serving web pages from data provided by the `gohbase` library. We feel that the hard-coded defaults for the timeouts mentioned in the title are far too long.

I am providing this pull request to give users of the library the ability to adjust the listed timeouts to suite their needs. If I have coded this correctly, the intention is to be completely backwards compatible with no public API changes, and the same default behavior as one would have experienced before the change. The timeouts listed above are exposed as `Option`s in the same way that other `Option`s are already exposed for the main hbase `client`.

The patch passes when calling `go test` on my workstation. Please review, and let me know if I can do anything to further assist, or if I should change anything to be more in-line with the project's goals.

Thanks!


travishegner:
I found the real command to run for executing tests correctly. Stand by for an update to the pull request.

travishegner:
Alright, PR now passes the tests correctly. Looking forward to any feedback.

travishegner:
Ugh, my apologies, looks like I modified more than intended with having the "simplify" flag set by default for  gofmt. I'll clean that up.

region/client.go
```diff
@@ -183,10 +195,14 @@ func (c *client) String() string {
 }
 
 func (c *client) inFlightUp() {
+	if c.readTimeout == 0 {
```
timoha: I think this should be set in region/client.New()

rpc.go
```diff
@@ -597,10 +592,17 @@ func (c *client) establishRegionClient(reg hrpc.RegionInfo,
...
 	}
-	clientCtx, cancel := context.WithTimeout(reg.Context(), regionLookupTimeout)
+	clientCtx, cancel := context.WithTimeout(reg.Context(), c.regionLookupTimeout)
 	defer cancel()
-	return region.NewClient(clientCtx, addr, clientType,
+
+	r, err := region.NewClient(clientCtx, addr, clientType,
 		c.rpcQueueSize, c.flushInterval, c.effectiveUser)
+
+	if err == nil {
```
timoha: I think you should just pass it as an argument to NewClient

hrpc/call.go
```diff
@@ -43,6 +44,8 @@ type RegionClient interface {
 	Addr() string
 	QueueRPC(Call)
 	String() string
+	SetReadTimeout(time.Duration)
```
timoha: there's probably no need for these functions in an interface, just pass the ReadTimeout as an argument to NewClient

zk/client.go
```diff
@@ -55,10 +53,12 @@ const (
 // Client is an interface of client that retrieves meta infomation from zookeeper
 type Client interface {
 	LocateResource(ResourceName) (string, error)
+	SetTimeout(time.Duration)
```
timoha: same thing, just pass it as an argument to NewClient

zk/client.go
```diff
@@ -68,9 +68,20 @@ func NewClient(zkquorum string) Client {
...
+}
+
+func (c *client) GetTimeout(t time.Duration) {
+	c.sessionTimeout = t
+}
+
 // LocateResource returns address of the server for the specified resource.
 func (c *client) LocateResource(resource ResourceName) (string, error) {
-	conn, _, err := zk.Connect(c.zks, time.Duration(sessionTimeout)*time.Second)
+	if c.sessionTimeout == 0 {
```
timoha: call this in NewClient

travishegner:
Perfect, I was hoping you would suggest that as it's much cleaner. I originally started implementing it that way, and then feared that there may be some direct uses of your zk/client and region/client in the wild that you wanted to maintain compatibility for.

timoha:
Good point, but not that I know of. The reason those are not internal is because gomock doesn’t work as expected with internal packages. Maybe that has changed, need to give it a try

travishegner:
My only experience with gomock is what you've seen me fumble through here, so I can't comment intelligently on that.

However, I would recommend adding any generated code to a `.gitignore`, and having your CI environment generate it dynamically on each build. I initially missed the "DO NOT EDIT" comments at the top of the file, and didn't realize that they were automatically generated until I started researching how gomock worked. It's not a critical issue, but it helps keep the repo a little cleaner.

In any case, do you have any further comments on this patch? Would you prefer it rebased as a single commit?

Thanks for taking the time to review it.

timoha:
> However, I would recommend adding any generated code to a .gitignore, and having your CI environment generate it dynamically on each build. I initially missed the "DO NOT EDIT" comments at the top of the file, and didn't realize that they were automatically generated until I started researching how gomock worked. It's not a critical issue, but it helps keep the repo a little cleaner.

good idea, I'll keep it my TODO

> In any case, do you have any further comments on this patch? Would you prefer it rebased as a single commit?

rebasing as single commit would be great. we using gerrit internally, so usually our commits are self contained and history looks cleaner.

Also please sing the CLA if you haven't already: https://github.com/tsuna/gohbase#contributing.

travishegner:
OK, all changes are rebased to a single commit. I have already signed the CLA, so let me know if you need anything else.

Thanks!

timoha:
LGTM, thanks for contributing!

PR #70, opened 2017-09-26
Implement reverse scanner

dougklein:
HBase offers reverse scanning functionality which can be very useful with time series keys. 

Initial attempt did not change regions correctly, but after reviewing the java code for this, I think this is handling region changes in a similar way.

Included integration tests with the test table pre-split for my test data ranges.

### DIFF
--- a/hrpc/scan.go
+++ b/hrpc/scan.go
@@ -70,6 +70,7 @@ type Scan struct {
 
 	maxResultSize uint64
 	numberOfRows  uint32
+	reversed      bool
 
 	closeScanner        bool
 	allowPartialResults bool
@@ -88,6 +89,7 @@ func baseScan(ctx context.Context, table []byte,
 		scannerID:     math.MaxUint64,
 		maxResultSize: DefaultMaxResultSize,
 		numberOfRows:  DefaultNumberOfRows,
+		reversed:      false,
 	}
 	err := applyOptions(s, options...)
 	if err != nil {
@@ -184,6 +186,11 @@ func (s *Scan) AllowPartialResults() bool {
 	return s.allowPartialResults
 }
 
+// Reversed returns true if scanner scans in reverse.
+func (s *Scan) Reversed() bool {
+	return s.reversed
+}
+
 // ToProto converts this Scan into a protobuf message
 func (s *Scan) ToProto() proto.Message {
 	scan := &pb.ScanRequest{
@@ -224,6 +231,9 @@ func (s *Scan) ToProto() proto.Message {
 	}
 	if s.toTimestamp != MaxTimestamp {
 		scan.Scan.TimeRange.To = &s.toTimestamp
+	}
+	if s.reversed {
+		scan.Scan.Reversed = &s.reversed
 	}
 	scan.Scan.Filter = s.filter
 	return scan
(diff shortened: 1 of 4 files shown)


scanner.go
```diff
@@ -294,7 +294,28 @@ func (f *fetcher) update(resp *pb.ScanResponse, region hrpc.RegionInfo) {
...
+
+		// Reversed Scan
+		// return if we are at the end
+		if len(region.StartKey()) == 0 {
+			f.startRow = region.StartKey()
+			return
+		}
+
+		// create the nearest value lower than the current region startKey
+		// this won't work correctly on keys  / splits with embedded 0x0.
```
timoha: Not supporting embedded `0x0` is pretty big limitation IMHO.
Seems like it could be easily solved by doing something like this:
```
if tmp[len(tmp)-1] == 0x0 {
     f.startRow = tmp[:len(tmp)-1]
     return
}
```

dougklein:
Thank you for your feedback. I've rebased the library again and applied changes - including your suggestions from the last pull request.

scanner.go
```diff
@@ -319,11 +348,19 @@ func (f *fetcher) shouldClose(resp *pb.ScanResponse, region hrpc.RegionInfo) boo
...
 	// (3) because its stop_key is greater than or equal to the stop_key of this scanner,
 	// provided that (2) we're not trying to scan until the end of the table.
+	if !f.rpc.Reversed() {
+		return len(f.rpc.StopRow()) != 0 && // (2)
+			bytes.Compare(f.rpc.StopRow(), region.StopKey()) <= 0 // (3)
+	}
+	//  Reversed Scanner
 	return len(f.rpc.StopRow()) != 0 && // (2)
-		bytes.Compare(f.rpc.StopRow(), region.StopKey()) <= 0 // (3)
+		bytes.Compare(f.rpc.StopRow(), region.StartKey()) >= 0 // (3)}
```
timoha: we shouldn't stop if StopRow() of the scan is equal to StartKey() of the region because it's inclusive.

scanner.go
```diff
@@ -294,7 +294,36 @@ func (f *fetcher) update(resp *pb.ScanResponse, region hrpc.RegionInfo) {
...
+		// if last element is 0x0, just shorten the slice
+		if tmp[len(tmp)-1] == 0x0 {
+			f.startRow = tmp[:len(tmp)-1]
+			return
+		}
+
+		// otherwise lower the last element byte value by 1
+		// and pad with 0xff
+		tmp[len(tmp)-1] = tmp[len(tmp)-1] - 1
+		tmp = append(tmp, ([]byte{0xff, 0xff, 0xff, 0xff})...)
```
timoha: `[]byte{0xff, 0xff, 0xff, 0xff}` should be a constant.

hrpc/scan.go
```diff
@@ -304,3 +314,16 @@ func AllowPartialResults() func(Call) error {
 		return nil
 	}
 }
+
+// SetReversed is a Scan-only option which allows you to scan in reverse key order
+// To use it the startKey would be greater than the end key
+func SetReversed(reversed bool) func(Call) error {
```
timoha: just `Reversed(...)` is cleaner.

scanner.go
timoha: actually nevermind, StopRow isn't inclusive.

timoha:
@dougklein could you allow pushing to your fork (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)? I have addressed the comments myself and would like to push them to this PR.

dougklein:
Thank you for merging this.

PR #73, opened 2017-10-14
using recommended lowercase reference to logrus

netguy204:
Logrus authors recommend that references to the logrus package use only the lowercase variation. See https://github.com/sirupsen/logrus/blob/master/README.md

End user code will not build unless that code and all of its dependencies follow only a single logrus case convention.

### DIFF
--- a/admin_client.go
+++ b/admin_client.go
@@ -9,7 +9,7 @@ import (
 	"context"
 	"fmt"
 
-	log "github.com/Sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	"github.com/tsuna/gohbase/hrpc"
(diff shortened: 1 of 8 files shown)


netguy204:
Can someone with more experience with the CI system take a look a these build logs? I'm not sure if I caused or should be concerned about the various error-like lines in the build output.

timoha:
the reason it's failing is because you imports are out of order for the following files:
caches.go
client.go
rpc.go
scanner.go

Just run them again gofmt to correct them

timoha:
Thanks for contributing, in case you haven't done this yet, please sign the contributors agreement: https://github.com/tsuna/gohbase#contributing

PR #97, opened 2018-07-26
Added admin client support for listing table names

ozkatz:
This pull request adds a new admin client method named `ListTableNamesByNamespace` (same name as the HBase protobuf interface command). Given a context and a namespace, it will return a list of table names.

It is currently possible to achieve by scanning the `hbase:meta` table, but since it is keyed by both table and region, it doesn't scale well to large HBase installations where tables may have a very large amount of regions (for my particular case, this change reduced response time from multiple seconds to <1 second).


timoha:
implemented via https://github.com/tsuna/gohbase/pull/124

PR #120, opened 2019-07-22
Add snapshot creation and deletion

devoxel:
This commit introduces the ability to add and delete hbase snapshots via the
admin client.

Due to the snapshot API being slightly different to other admin calls, we introduce a new option SnapshotValidateInterval which allows us to know when the Snapshot has been created.

One slight oddity of behavior can come if SnapshotValidateInterval is long enough that a context deadline is hit during creation even though the snapshot was created. I decided to always delete in this case -- but perhaps it would be better to return with no error. Let me know what you think.


admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
 		}
 	}
 }
+
+// Snapshot creates a snapshot in HBase.
+//
+// Given that this operation takes some time to run, it's reccomended you
```
timoha: typo

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+				return fmt.Errorf("sendPRC returned not a IsSnapshotDoneResponse")
+
+			}
+
+			if *r.Done {
+				return nil
+			}
+		case <-ctx.Done():
+			t.SetContext(context.Background())
+			err := c.DeleteSnapshot(t)
```
timoha: Could we not do this but let the caller decide themselves what to do in case of a timeout. It's similar situation as with CreateTable and such

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+// is received, the client will not cleanup the snapshot you created.
+func (c *client) Snapshot(t *hrpc.Snapshot) error {
+	pbmsg, err := c.SendRPC(t)
+	if err != nil {
+		return err
+	}
+
+	_, ok := pbmsg.(*pb.SnapshotResponse)
+	if !ok {
+		return fmt.Errorf("sendPRC returned not a SnapshotResponse")
```
timoha: errors.New

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+		select {
+		case <-ticker.C:
+			pbmsgs, err := c.SendRPC(check)
+			if err != nil {
+				return err
+			}
+
+			r, ok := pbmsgs.(*pb.IsSnapshotDoneResponse)
+			if !ok {
+				return fmt.Errorf("sendPRC returned not a IsSnapshotDoneResponse")
```
timoha: errors.New

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+
+			}
+
+			if *r.Done {
```
timoha: r.GetDone() otherwise it can panic if Done is nil

hrpc/snapshot.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// CheckSnapshot represents a CheckSnapshot HBase call
+type CheckSnapshot struct {
+	*Snapshot
+}
+
+// NewCheckSnapshot creates a new CheckSnapshot request that will request a
+// new snapshot in HBase.
+func NewCheckSnapshot(t *Snapshot) *CheckSnapshot {
```
timoha: could we call it consistently with service name "IsSnapshotDone"? CheckSnapshot seems like a more generic name which assume that it validates that snapshot is intact.

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+	rt := hrpc.NewDeleteSnapshot(t)
+
+	pbmsg, err := c.SendRPC(rt)
+	if err != nil {
+		return err
+	}
+
+	_, ok := pbmsg.(*pb.DeleteSnapshotResponse)
+	if !ok {
+		return fmt.Errorf("sendPRC returned not a DeleteSnapshotResponse")
```
timoha: errors.New

admin_client.go
```diff
@@ -23,6 +24,8 @@ type AdminClient interface {
 	DeleteTable(t *hrpc.DeleteTable) error
 	EnableTable(t *hrpc.EnableTable) error
 	DisableTable(t *hrpc.DisableTable) error
+	Snapshot(t *hrpc.Snapshot) error
```
timoha: "CreateSnapshot" sound more consistent maybe?

admin_client.go
```diff
@@ -154,3 +158,75 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+		return err
+	}
+
+	_, ok := pbmsg.(*pb.DeleteSnapshotResponse)
+	if !ok {
+		return fmt.Errorf("sendPRC returned not a DeleteSnapshotResponse")
+
+	}
+
+	// TODO: It might be worth ensuring the snapshot no longer exists.
```
timoha: It looks like DeleteSnapshotRequest is blocking and once it returns the response, the snapshot directory is removed (because they call post procedure hook): https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/master/snapshot/SnapshotManager.java#L340

so you don't need to check

devoxel:
All comments addressed. I force pushed to keep it as a single commit -- hope that's good for you.

admin_client.go
```diff
@@ -154,3 +158,62 @@ func (c *client) checkProcedureWithBackoff(ctx context.Context, procID uint64) e
...
+	if err != nil {
+		return err
+	}
+
+	_, ok := pbmsg.(*pb.SnapshotResponse)
+	if !ok {
+		return errors.New("sendPRC returned not a SnapshotResponse")
+	}
+
+	ticker := time.NewTicker(c.snapValidateInterval)
```
timoha: need to stop the ticker: https://golang.org/pkg/time/#Ticker.Stop

client.go
```diff
@@ -197,6 +201,14 @@ func FlushInterval(interval time.Duration) Option {
 	}
 }
 
+// SnapshotValidateInterval will return an option that will set the interval
+// at which RPC calls are made when validating that a snapshot has finished.
+func SnapshotValidateInterval(interval time.Duration) Option {
```
timoha: do we really need this option? it seems like we should just use some sane default and not give extra burden on the client to even consider changing it.

hrpc/snapshot.go
```diff
@@ -0,0 +1,167 @@
+// Copyright (C) 2015  The GoHBase Authors.  All rights reserved.
```
timoha: 2019

client.go
devoxel: Removed the option. I'm not sure what a sane option is. I chose half a second here but I'm happy to change it.

PR #124, opened 2019-08-21
hrpc, gohbase: Add table listing to admin client

devoxel:
Fairly straightforward.

### DIFF
--- a/admin_client.go
+++ b/admin_client.go
@@ -35,6 +35,7 @@ type AdminClient interface {
 	ListSnapshots(t *hrpc.ListSnapshots) ([]*pb.SnapshotDescription, error)
 	RestoreSnapshot(t *hrpc.Snapshot) error
 	ClusterStatus() (*pb.ClusterStatus, error)
+	ListTableNames(t *hrpc.ListTableNames) ([]*pb.TableName, error)
 }
 
 // NewAdminClient creates an admin HBase client.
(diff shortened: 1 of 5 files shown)


hrpc/list.go
```diff
@@ -0,0 +1,50 @@
...
+type GetTableNames struct {
+	base
+	regex            string
+	includeSysTables bool
+	namespace        string
+}
+
+// NewGetTableNames creates a new GetTableNames request that will list
+// tables in hbase
+func NewGetTableNames(ctx context.Context, table string, regex string) *GetTableNames {
```
timoha: do you mean "namespace" instead of "table"? Otherwise I don't see where you pass namespace, maybe add a test for that.

admin_client.go
```diff
@@ -35,6 +35,7 @@ type AdminClient interface {
 	ListSnapshots(t *hrpc.ListSnapshots) ([]*pb.SnapshotDescription, error)
 	RestoreSnapshot(t *hrpc.Snapshot) error
 	ClusterStatus() (*pb.ClusterStatus, error)
+	GetTableNames(t *hrpc.GetTableNames) ([]*pb.TableName, error)
```
timoha: Could we change it be as "ListTableNames" just to be consistent with java api?

hrpc/list.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/golang/protobuf/proto"
+	"github.com/tsuna/gohbase/pb"
+)
+
+// GetTableNames models a GetTableNames pb call
+type GetTableNames struct {
+	base
+	regex            string
+	includeSysTables bool
```
timoha: how do you pass this?

hrpc/list.go
devoxel: These were a TODO in the previous commit. Modified them to use the option pattern and added tests to ensure that works in the integration test.

admin_client.go
devoxel: Done

hrpc/list.go
devoxel: See above comment

PR #140, opened 2020-05-22
Add snappy compression between client and regionserver

timoha:
- Implements cellblocks serialization for mutate and multi
- Adds snappy compression and decompression for cellblocks

### DIFF
--- a/region/client.go
+++ b/region/client.go
@@ -190,6 +190,9 @@ type client struct {
 
 	// readTimeout is the maximum amount of time to wait for regionserver reply
 	readTimeout time.Duration
+
+	// compressor for cellblocks. if nil, then no compression
+	compressor *compressor
 }
 
 // QueueRPC will add an rpc call to the queue for processing by the writer goroutine
@@ -531,6 +534,13 @@ func (c *client) receive() (err error) {
 	}
 	if d, ok := rpc.(canDeserializeCellBlocks); cellsLen > 0 && ok {
 		b := b[size-cellsLen:]
+		if c.compressor != nil {
+			b, err = c.compressor.decompressCellblocks(b)
+			if err != nil {
+				err = RetryableError{fmt.Errorf("failed to decompress the response: %s", err)}
+				return
+			}
+		}
 		var nread uint32
 		nread, err = d.DeserializeCellBlocks(response, b)
 		if err != nil {
@@ -538,7 +548,8 @@ func (c *client) receive() (err error) {
 			return
 		}
 		if int(nread) < len(b) {
-			err = RetryableError{fmt.Errorf("short read: buffer len %d, read %d", len(b), nread)}
+			err = RetryableError{
+				fmt.Errorf("short read: buffer length %d, read %d", len(b), nread)}
 			return
 		}
 	}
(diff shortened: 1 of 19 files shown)


region/client.go
```diff
@@ -621,11 +642,20 @@ func (c *client) send(rpc hrpc.Call) (uint32, error) {
...
 
-	if err := c.write(b); err != nil {
+	if cellblocks != nil {
+		bfs := append(net.Buffers{b}, cellblocks...)
+		_, err = bfs.WriteTo(c.conn)
+	} else {
+		err = c.write(b)
+	}
+	// free header/request buffer
+	freeBuffer(b)
```
tsuna, line 678: This used to be done via a `defer`, shouldn't it still be?

region/client_test.go
```diff
@@ -942,6 +958,131 @@ func TestSanity(t *testing.T) {
...
+		conn:          mockConn,
+		rpcs:          make(chan hrpc.Call),
+		done:          make(chan struct{}),
+		sent:          make(map[uint32]hrpc.Call),
+		rpcQueueSize:  1, // size one to skip sendBatch
+		flushInterval: 1000 * time.Second,
+		compressor:    &compressor{Codec: mockCodec{}},
+	}
+
+	var wg sync.WaitGroup
```
tsuna, line 975: Nit: Only ever `Add()`ed to once, you could just use channel that gets closed.

region/compressor.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/tsuna/gohbase/compression"
+)
+
+type compressor struct {
+	compression.Codec
+}
+
+func min(x, y uint32) int {
+	if x > y {
```
tsuna, line 21: that's cute, implementing `min()` with a greater than check :)

region/compressor.go
```diff
@@ -0,0 +1,106 @@
...
+		b, chunkLen = c.Encode(uncompressedBuffer[:n], b)
+
+		// write the chunk length
+		binary.BigEndian.PutUint32(b[lenOffset:], chunkLen)
+
+		if err != nil {
+			break
+		}
+	}
+	freeBuffer(uncompressedBuffer)
```
tsuna, line 63: shouldn't this be called via `defer`?

region/compressor.go
```diff
@@ -0,0 +1,106 @@
...
+		// grow for chunk length
+		lenOffset = len(b)
+		b = growBuffer(b, 4)
+
+		b, chunkLen = c.Encode(uncompressedBuffer[:n], b)
+
+		// write the chunk length
+		binary.BigEndian.PutUint32(b[lenOffset:], chunkLen)
+
+		if err != nil {
```
tsuna, line 59: is any error expected here other than EOF?

region/compressor.go
timoha: Brain works in mysterious ways :D

region/client_test.go
timoha: used on line 1046 as well

region/compressor.go
timoha: Done

region/compressor.go
timoha: Done, panicing on any other error

PR #142, opened 2020-07-08
Fix problems with closing region scanners.

clarax:
This PR fixes two bugs on closing region scanners:

1. During scanning multiple regions, GoHBase requests a new region scanner on server side  when current region is exhausted, without closing the current one on server side. The fix uses the scanner id to close (https://issues.apache.org/jira/browse/HBASE-18000)

2. At the end of scan, gohbase always calls to close the scanner on server side. Some scan requests, such as meta lookups, has closeScanner flag set so server auto close them after the first response. This redundant call also causes the server( without the fix in https://issues.apache.org/jira/browse/HBASE-17489 ) to throw errors. hbase meta lookup request is called for every table access so this behavior almost doubles the load on hbase meta. For older version of hhase, it also spams the error log. The fix doesn't open a new scanner on client side so the call will be skipped.

The PR is tested and verified in large-scale production cluster.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -156,7 +156,7 @@ func TestGet(t *testing.T) {
 	}
 
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	err := insertKeyValue(c, key, "cf", val)
 	if err != nil {
 		t.Errorf("Put returned an error: %v", err)
@@ -198,7 +198,7 @@ func TestGet(t *testing.T) {
 func TestGetDoesntExist(t *testing.T) {
 	key := "row1.5"
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	headers := map[string][]string{"cf": nil}
 	get, err := hrpc.NewGetStr(context.Background(), table, key, hrpc.Families(headers))
 	rsp, err := c.Get(get)
@@ -219,6 +219,7 @@ func TestGetDoesntExist(t *testing.T) {
 
 func TestMutateGetTableNotFound(t *testing.T) {
 	c := gohbase.NewClient(*host)
+	defer c.Close()
 
 	key := "whatever"
 	table := "NonExistentTable"
@@ -246,7 +247,7 @@ func TestMutateGetTableNotFound(t *testing.T) {
 func TestGetBadColumnFamily(t *testing.T) {
 	key := "row1.625"
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	err := insertKeyValue(c, key, "cf", []byte("Bad!"))
 	if err != nil {
 		t.Errorf("Put returned an error: %v", err)
@@ -265,7 +266,7 @@ func TestGetBadColumnFamily(t *testing.T) {
 func TestGetMultipleCells(t *testing.T) {
 	key := "row1.75"
 	c := gohbase.NewClient(*host, gohbase.FlushInterval(time.Millisecond*2))
-
+	defer c.Close()
 	err := insertKeyValue(c, key, "cf", []byte("cf"))
 	if err != nil {
 		t.Errorf("Put returned an error: %v", err)
@@ -293,6 +294,7 @@ func TestGetMultipleCells(t *testing.T) {
 
 func TestGetNonDefaultNamespace(t *testing.T) {
 	c := gohbase.NewClient(*host)
+	defer c.Close()
 
 	get, err := hrpc.NewGetStr(context.Background(), "hbase:namespace", "default")
 	if err != nil {
@@ -314,7 +316,7 @@ func TestPut(t *testing.T) {
 		t.Fatal("Host is not set!")
 	}
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	putRequest, err := hrpc.NewPutStr(context.Background(), table, key, values)
 	if err != nil {
 		t.Errorf("NewPutStr returned an error: %v", err)
@@ -339,7 +341,7 @@ func TestPutMultipleCells(t *testing.T) {
 	values["cf"]["b"] = []byte("b")
 	values["cf2"]["a"] = []byte("a")
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	putRequest, err := hrpc.NewPutStr(context.Background(), table, key, values)
 	_, err = c.Put(putRequest)
 	if err != nil {
@@ -369,7 +371,7 @@ func TestMultiplePutsGetsSequentially(t *testing.T) {
 	keyPrefix := "row3"
 	headers := map[string][]string{"cf": nil}
 	c := gohbase.NewClient(*host, gohbase.FlushInterval(time.Millisecond))
-
+	defer c.Close()
 	err := performNPuts(keyPrefix, num_ops)
 	if err != nil {
 		t.Errorf("Put returned an error: %v", err)
@@ -394,6 +396,7 @@ func TestMultiplePutsGetsSequentially(t *testing.T) {
 
 func TestMultiplePutsGetsParallel(t *testing.T) {
 	c := gohbase.NewClient(*host)
+	defer c.Close()
 
 	const n = 1000
 	var wg sync.WaitGroup
@@ -442,7 +445,7 @@ func TestMultiplePutsGetsParallel(t *testing.T) {
 func TestTimestampIncreasing(t *testing.T) {
 	key := "row4"
 	c := gohbase.NewClient(*host)
-
+	defer c.Close()
 	var oldTime uint64 = 0
 	headers := map[string][]string{"cf": nil}
 	for i := 0; i < 10; i++ {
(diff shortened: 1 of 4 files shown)


scanner.go
```diff
@@ -278,14 +249,26 @@ func (s *scanner) update(resp *pb.ScanResponse, region hrpc.RegionInfo) {
...
 }
 
-// shouldClose check if this scanner should be closed and should stop fetching new results
-func (s *scanner) shouldClose(resp *pb.ScanResponse, region hrpc.RegionInfo) bool {
+func (s *scanner) Close() error {
+	if s.closed {
+		return errors.New("scanner has already been closed")
+	}
+	s.closed = true
+	if !s.isRegionScannerClosed() {
```
timoha, line 257: is this check needed as isRegionScannerClosed is already idempotent?

scanner.go
```diff
@@ -241,13 +213,12 @@ func (s *scanner) request() (*pb.ScanResponse, hrpc.RegionInfo, error) {
...
-			s.scannerID = resp.GetScannerId()
-		}
-	} else {
+	if s.isRegionScannerClosed() && resp.ScannerId != nil {
+		s.openRegionScanner(resp.GetScannerId())
+	}
+	if !resp.GetMoreResultsInRegion() {
 		// we are done with this region, prepare scan for next region
-		s.scannerID = noScannerID
+		s.closeRegionScanner()
```
timoha, line 221: so does this mean that for every scanned region, it's going to send an RPC to regionserver to close the scanner and regionserver doesn't automatically close those? I guess my understanding was wrong when I implemented this.

scanner.go
```diff
@@ -309,3 +292,40 @@ func (s *scanner) shouldClose(resp *pb.ScanResponse, region hrpc.RegionInfo) boo
...
+		panic(fmt.Sprintf("should not happen: previous region scanner was not closed"))
+	}
+	s.curRegionScannerID = scannerId
+}
+
+func (s *scanner) closeRegionScanner() {
+	if s.isRegionScannerClosed() {
+		return
+	}
+	if !s.rpc.IsClosing() {
```
timoha, line 311: nice catch on this one

scanner.go
```diff
@@ -241,13 +213,12 @@ func (s *scanner) request() (*pb.ScanResponse, hrpc.RegionInfo, error) {
 
 // update updates the scanner for the next scan request
 func (s *scanner) update(resp *pb.ScanResponse, region hrpc.RegionInfo) {
-	if resp.GetMoreResultsInRegion() {
-		if resp.ScannerId != nil {
-			s.scannerID = resp.GetScannerId()
-		}
-	} else {
+	if s.isRegionScannerClosed() && resp.ScannerId != nil {
```
timoha, line 216: so we don't want to update scanner ID every time? I guess we don't want to rely on HBASE-18000 as it's not desired way?

scanner_test.go
```diff
@@ -9,18 +9,17 @@ import (
...
-	"testing"
-
 	"github.com/golang/mock/gomock"
 	"github.com/tsuna/gohbase/hrpc"
 	"github.com/tsuna/gohbase/pb"
 	"github.com/tsuna/gohbase/region"
 	"github.com/tsuna/gohbase/test"
 	"github.com/tsuna/gohbase/test/mock"
 	"google.golang.org/protobuf/proto"
+	"io"
```
timoha, line 19: could you move these back up as once we enable linter, it will complain

scanner_test.go
```diff
@@ -93,11 +92,27 @@ func dup(a []*pb.Result) []*pb.Result {
...
+		group.Done()
+	}).Return(&pb.ScanResponse{}, nil).Times(1)
+	return nil
+}
 func TestScanner(t *testing.T) {
 	ctrl := test.NewController(t)
 	defer ctrl.Finish()
 	c := mock.NewMockRPCClient(ctrl)
 
+	var wg sync.WaitGroup
```
timoha, line 112: that's a bit of an odd way to use wait group. The test would stall if the SendRPC from Close isn't called and it would be better to fail the test instead. The fact that you've setup and expect call should already achieve that, so you don't need the wait group here.

scanner_test.go
```diff
@@ -93,11 +92,27 @@ func dup(a []*pb.Result) []*pb.Result {
 	return b
 }
 
+func testCallClose(scan *hrpc.Scan, c *mock.MockRPCClient, scannerID uint64,
```
timoha, line 95: you can pass t *testing.T to this function and call t.Helper() at the top of the body. Then handle error from NewScanRange within the function so that you don't need to have those "if" everywhere you call it.

integration_test.go
```diff
@@ -156,7 +156,7 @@ func TestGet(t *testing.T) {
 	}
 
 	c := gohbase.NewClient(*host)
-	defer c.Close()
```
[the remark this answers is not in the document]
clarax: The documentation on the interface says the client will close automatically, which wasn't true with the bug. With the bug, close is always called but only tries to the last region scanner again, leaving scanners in previous scanners open on server side. Even with a redundant catch-all close call, with the bug, this With the fix, there is no need for this call, doing exactly what the documentation says. And I think it is better to show a good example in tests conforming to the documentation.

scanner.go
clarax: The server would close them after timing out, which is 5 min later by default, wasting resources and throwing warnings. For a cluster with heavy load of small scans, it is not scalable.

integration_test.go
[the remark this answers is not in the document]
timoha: I see the confusion here. The documentation of the Scanner.Close (https://godoc.org/github.com/tsuna/gohbase/hrpc#Scanner) says that it's closed automatically. However, in this case you are dealing with/closing the gohbase client (https://godoc.org/github.com/tsuna/gohbase#Client)

scanner.go
clarax: client should do best effort to call close as soon as done. Server closing scanner is the last resort.

timoha: Wondering if regionserver could be a bit smarter here as it knows that the scanner for the region has been exhausted and close it automatically rather than waiting for timeout (but that's a server side change). Thanks for explanation.

scanner.go
clarax: This is to make sure we have closed the scanner on last region and this is for a new region. As client, we cannot update scanner id if not returned from server.

scanner_test.go
clarax: Hmm, they were moved by gofmt. If I move back, make fmt complains.

scanner_test.go
clarax: I may get a missing calls error if I don't set up the wait. I put it in to avoid flaky tests.

scanner_test.go
timoha: that's weird, which version of go are you using? you also need to add a blank line between them

scanner.go
clarax: I also removed the check for resp.GetMoreResultsInRegion() because we need the scanner id for the call to close region scanner.

scanner_test.go
clarax: 1.14 

I can add a blank line.

integration_test.go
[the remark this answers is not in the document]
clarax: That was how the application developers interpreted. But anyway, the implementation does automatically close the client in all cases I examined. No close from user code needed.

scanner_test.go
timoha: uh I think this is do to the fact that we are calling the closing SendRPC in a goroutine. I never really liked that and it does make the tests less predictable and can potentially end up not closing the scanner in production (for example if program execution has finished). This can be resolved in another PR if you want to work on it.

timoha: or can just drop `go` part in this review as it should be trivial (ignoring the error). The clients can themselves call Close() in a goroutine  if the really want it to be done asynchnously

clarax: Yes. I understood why I get missing call errors and then put in the wait for test. If the test stalls, it is a test failure. We do expect it to be called.  BTWI don't think the go routine will cause leak if a program is gracefully shutdown. And it is OK to have some left for server to close in that exceptional case.

scanner.go
clarax: will remove.

integration_test.go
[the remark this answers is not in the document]
clarax: Now I recall why I remove them in the first place. Those calls generate a couple unclosed region scanners for two test cases: TestScannerClose and Test ScannerContextCancel. But we can look at them as a separate issue.

scanner_test.go
clarax: Thank you. that is helpful.

scanner_test.go
clarax: added

scanner_test.go
clarax: And there are other tests using wait group too.

clarax:
Thank you for the approval @timoha could you merge it?

PR #145, opened 2020-08-20
hrpc: expose CacheBlocks option for scan/get queries

asetty:
Client.proto already has this defined as `cache_blocks` in the models for `Get` and `Scan`. This allows users of the hrpc to set the attribute per request.


asetty:
moved to https://github.com/tsuna/gohbase/pull/147 to change branch the pull request is from

Issue #141, opened 2020-05-25
Kerberos Support?


coderw184:
Hi,

Does this library support Kerberos and if so is its use documented somewhere?

Thanks

Issue #95, opened 2018-06-28
Kerberos Support?


HariSekhon:
Hi,

Does this library support Kerberos and if so is it's use documented somewhere?

Thanks

Hari

IDerr:
Perhaps it's possible to integrate kerberos with this library : 
https://github.com/jcmturner/gokrb5

vderic:
+1

3rduncle:
+1

Issue #71, opened 2017-09-27
kerberos support?


chy373:
how can I work with a hbase cluster with kerberos?

niklausxiao:
seems no kerberos support yet.

I searched lots of golang implementation of hadoop client, include hdfs,yarn, hbase, kafka, none of them support kerberos. :(

timoha:
gohbase doesn't support any authentication methods at the moment as we haven't had a need for it ourselves. If you think this could be useful and are interested to help out, you are welcome to send a PR

PR #148, opened 2020-09-01
Kerberos proof of concept

josiahg:
While this code works, this is not yet a fully baked solution. Creating a PR for discussion on the approach and how Kerberos support should best be integrated into gohbase.

Kerberos support has been requested several times on this project -  #71 #95 #141 - I needed it as well, so I took a shot at it.

First, what IS here:
- Basic Kerberos authentication support for GSSAPI applications on a user account that has already been granted a ticket through `kinit`.
- A faithful representation of the SASL handshake flow used by HBase, pulled from the Java RPC Client code.

What's NOT here:
- Support for MD5-DIGEST Kerberos authentication
- Support for on-going encrypted transport 
- Support for passing custom Keytabs, `krb5.conf` or other configuration files
- Tests 😀 

What's here, but maybe shouldn't be:
- Relies on [beltran/gosasl](https://github.com/beltran/gosasl) - this in turn relies on C libraries to handle the SASL bits. I would prefer to use [jcmturner/gokrb5](https://github.com/jcmturner/gokrb5) as this is a pure Go implementation. However, it is missing some of the SASL bits and I will need to reimplement those.
- Due to the above dependency, you must have some libraries install and run/build with `-tags kerberos` to make things work

Open questions (not by any means an exhaustive list):
- Will the gohbase project accept a Kerberos solution that is not pure go? IE: if the above dependency chain with gosasl cannot be avoided
- What are the minimum additional features beyond what is currently in this PR that must be included for this to be merged?
- What is the correct mechanism for adding this to the workflow? Currently I've made "SIMPLE" auth the default, and added a config option for setting auth to "KERBEROS"

Would love any feedback or thoughts from the team or community!

Finally, to test it - 

Assumptions: 
- This is done on a machine that can `kinit` to the KDC with a user with HBase permissions)
- You have followed the install instructions for `gosasl`:

To add Kerberos support gosasl requires header files to build against the GSSAPI C library. They can be installed with:

Ubuntu: `sudo apt-get install libkrb5-dev`
MacOS: `brew install homebrew/dupes/heimdal --without-x11`
Debian: `yum install -y krb5-devel`

First, pull in the fork's branch so your app can use it:

```
go get github.com/tsuna/gohbase
cd $GOPATH/src/github.com/tsuna/gohbase
git remote add fork https://github.com/josiahg/gohbase
git fetch fork
git checkout kerb
```

Everything else is the same as a standard gohbase application, with the addition of the auth option and ensuring the effective user is set to the Kerberos principle you will authenticate as:

```
...
auth := gohbase.Auth("KERBEROS")
user := gohbase.EffectiveUser("user@EXAMPLE.COM")
options := []gohbase.Option{auth, user}
client := gohbase.NewClient(url, options...)
...
```

Ensure you you have `kinit` as the EffectiveUser set above before running your application.

Finally, due to the dependencies in `gosasl` you must currently run with `-tags kerberos`:

```
go run -tags kerberos main.go
```

or 
```
go build -tags kerberos
```


microeastcowboy:
when zk with kerberos security enabled，it will fail. 
do we have any plan support for zk sasl auth ?

zhuliquan:
@tsuna We've already verified this code in a real online environment that it's real and valid, and hopefully merge this code

JcKendo:
i tested on HBase Version 2.4.16, it does not work -_- !

PR #150, opened 2020-10-06
Don't return error on subsequent closing of scanner

timoha:
We are calling Close() on scanner internally when there's no more data,
    the client might also additionally call Close() themselves causing
    unexpected for them error.


tsuna (approved):
Yeah Close() should really be idempotent. We should probably change the `Scanner` interface so that the godoc specifically mentions that it's okay to close an already closed scanner.

timoha:
> We should probably change the Scanner interface so that the godoc specifically mentions that it's okay to close an already closed scanner.
Done.

PR #152, opened 2020-10-30
New Admin APIs

timoha:
- Add `SetBalancer` to enable/disable balancer
- Add `MoveRegion` to move regions

### DIFF
--- a/hrpc/move.go
+++ b/hrpc/move.go
@@ -1,0 +1,92 @@
+// Copyright (C) 2020  The GoHBase Authors.  All rights reserved.
+// This file is part of GoHBase.
+// Use of this source code is governed by the Apache License 2.0
+// that can be found in the COPYING file.
+
+package hrpc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/tsuna/gohbase/pb"
+	"google.golang.org/protobuf/proto"
+)
+
+// MoveRegion allows to move region to a different RegionServer.
+type MoveRegion struct {
+	base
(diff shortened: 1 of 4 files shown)


hrpc/move.go
```diff
@@ -0,0 +1,89 @@
...
+
+// WithDestinationRegionServer specifies destination RegionServer for MoveReqion request
+// A server name is its host, port plus startcode: host187.example.com,60020,1289493121758
+func WithDestinationRegionServer(serverName string) func(Call) error {
+	return func(c Call) error {
+		mr, ok := c.(*MoveRegion)
+		if !ok {
+			return errors.New("WithDestinationRegionServer option can only be used with MoveRegion")
+		}
+		out := strings.SplitN(serverName, ",", 3)
```
tsuna, line 33: Need to check that `len(out)` is 3. `SplitN()` returns _at most_ N substrings, but could return 0 or 1 or 2.

tsuna (changes requested):

hrpc/move.go
timoha: Done

hrpc/move.go
```diff
@@ -0,0 +1,92 @@
...
+		if len(out) != 3 {
+			return errors.New("invalid server name, needs to be of format <host>,<port><startcode>")
```
tsuna, line 35: Missing a comma after <port>

tsuna (changes requested):

hrpc/move.go
timoha: Done

tsuna (approved):

PR #163, opened 2021-03-30
fix isRegionEstablished blocking

MAGE001:
If region server down for some time, and then comes back later. The client would not recreate a new connection but keep block sending in isRegionEstablished.


moredure:
@timoha , @tsuna why not to apply this change?

PR #164, opened 2021-05-18
Add mutex to limit the number of goroutine to lookup meta table

YutSean:
In our production environment, when a table contains a large quantity of regions (more than 1000 regions), the client will send too many scan requests to meta regionserver when the local cache is invalid. This results in a rapid heavy pressure to the meta regionserver and the meta regionserver may be dead. To protected the meta regionserver, add a mutex to limit the lookup of meta table will improve the stability of the whole cluster.


tsuna:
This limits the concurrency to 1 though, maybe we should use a semaphore to allow multiple concurrent lookups to proceed? Do you have any metrics to share in terms of the performance impact this change has?

YutSean:
I checked the HBase community Java Client implementation. It just use a lock to ensure only one thread check meta info at one time and in our usage, only one goroutine to check meta info is not noticeable by the users. Btw, meta lookup is a very fast process, use mutex can protect the regionserver where the meta region located.

PR #170, opened 2021-10-19
Added tracing and metrics

warrenhodg:
This PR adds Opentelemetry tracing and Prometheus metrics.


observability/observability.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2021  The GoHBase Authors.  All rights reserved.
+// This file is part of GoHBase.
+// Use of this source code is governed by the Apache License 2.0
+// that can be found in the COPYING file.
+
+package observability
```
tsuna, line 6: Could we move this under gohbase/internal/observability instead, to remove it from the public API of GoHBase?

region/client.go
```diff
@@ -319,24 +324,49 @@ func (c *client) unregisterRPC(id uint32) hrpc.Call {
...
 			}).Debug("flushing MultiRequest")
 		}
+
+		flushReasonCount.With(prometheus.Labels{
+			"reason": reason,
+		}).Inc()
+
+		m.addSendEventsToCallSpans()
+
+		callCount := m.callCount()
```
tsuna, line 355: Nit: this variable doesn't seem useful, maybe just call m.callCount() directly on line 358?

region/client.go
```diff
@@ -408,7 +441,23 @@ func returnResult(c hrpc.Call, msg proto.Message, err error) {
 	}
 }
 
-func (c *client) trySend(rpc hrpc.Call) error {
+func (c *client) trySend(rpc hrpc.Call) (err error) {
+	// Replace the rpc's context with one that includes a
+	// tracing span, but restore it when returning from this
+	// function so that spans don't wierdly cascade when they
```
tsuna, line 447: tyop: weirdly

region/client.go
```diff
@@ -319,24 +324,49 @@ func (c *client) unregisterRPC(id uint32) hrpc.Call {
 }
 
 func (c *client) processRPCs() {
+	spCtx, sp := observability.StartSpan(context.Background(), "region.processRPCs")
+	defer func() {
+		// Ensure that the last span created is ended to
+		// prevent memory leaks
```
tsuna, line 330: We also need to take care of this when we cancel all in-flight RPCs due to an error in failSentRPCs() above on line 306.

PR #173, opened 2021-11-17
ci: replace travis with github actions

dethi:


dethi:
Nice, it works. It even detected a race conditiong 😨

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup | Download HBase
+        run: |
+          ver=2.4.3 &&
```
tsuna, line 17: Make that 2.4.8?

hrpc/call.go
```diff
@@ -110,7 +111,9 @@ type RPCResult struct {
 }
 
 type base struct {
-	ctx     context.Context
+	ctxM sync.Mutex
```
tsuna, line 114: This is a symptom of the problem Warren raised when he added tracing support. We shouldn't store the context in the RPC object, we should pass it along, and when we derive a new context due to tracing we just pass along the derived context.

Adding this here is fine for now because it fixes a real bug, but the proper fix would be to remove context from this struct altogether. I filed #174 to track this.

tsuna (approved):

PR #177, opened 2021-12-06
admin api create/delete/disable/enable support tablename like ns:name

xfan001:
create/delete/disable/enable action support set namespace

### DIFF
--- a/hrpc/create.go
+++ b/hrpc/create.go
@@ -6,6 +6,7 @@
 package hrpc
 
 import (
+	"bytes"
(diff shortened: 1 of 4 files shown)


dethi (changes requested):
It would also make sense if the namespace was saved in the `base` struct (see `hrpc/call.go`). This would mean that the extraction happens in the various `NewX(...)` function and that we can add a `Namespace() []byte` function to the `base` struct (like we already have `Table() []byte`)

PR #178, opened 2021-12-06
tracing: indicate whether traces are sampled

warrenhodg:
In the case where trace sampling is applied, it could be that events are added with traceIDs that will not be readable.
This change simply indicates this state.


dethi:
Wouldn't it be better if it was the other way? (indicating when it's sampled)

tsuna (approved):

PR #180, opened 2022-01-27
Allow specification of consistency with Get and Scan

bwesterb:
https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.2.9/bk_hadoop-ha/content/ha-hbase-timeline-consistency.html

### DIFF
--- a/hrpc/get.go
+++ b/hrpc/get.go
@@ -113,6 +113,9 @@ func (g *Get) ToProto() proto.Message {
 	if g.cacheBlocks != DefaultCacheBlocks {
 		get.Get.CacheBlocks = &g.cacheBlocks
 	}
+	if g.consistency != DefaultConsistency {
+		get.Get.Consistency = g.consistency.toProto()
+	}
 	get.Get.Filter = g.filter
 	return get
 }
--- a/hrpc/call.go
+++ b/hrpc/call.go
@@ -101,6 +101,7 @@ type hasQueryOptions interface {
 	setMaxResultsPerColumnFamily(maxresults uint32)
 	setResultOffset(offset uint32)
 	setCacheBlocks(cacheBlocks bool)
+	setConsistency(consistency ConsistencyType)
 }
 
 // RPCResult is struct that will contain both the resulting message from an RPC
--- a/hrpc/query.go
+++ b/hrpc/query.go
@@ -24,6 +24,40 @@ type baseQuery struct {
 	storeLimit    uint32
 	storeOffset   uint32
 	cacheBlocks   bool
+	consistency   ConsistencyType
+}
+
+// ConsistencyType is used to specify the required consistency of data
+//
+// See https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.2.9/bk_hadoop-ha/
+//          content/ha-hbase-timeline-consistency.html
+type ConsistencyType int
+
+const (
+	// Use HBase's default
+	DefaultConsistency ConsistencyType = iota
+
+	// Guarantees that the client receives the latest data.
+	StrongConsistency
+
+	// Client might receive stale data (indicated by the Stale field), but
+	// the data that is received was valid at a given point of time.
+	TimelineConsistency
+)
+
+func (c ConsistencyType) toProto() (ret *pb.Consistency) {
+	ret = new(pb.Consistency)
+	switch c {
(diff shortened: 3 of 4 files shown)


hrpc/get.go
```diff
@@ -18,8 +19,41 @@ type Get struct {
...
+
+	// Guarantees that the client receives the latest data.
+	StrongConsistency
+
+	// Client might receive stale data (indicated by the Stale field), but
+	// the data that is received was valid at a given point of time.
+	TimelineConsistency
+)
+
+func (c QueryConsistency) ToProto() (ret *pb.Consistency) {
```
Lekensteyn, line 44: Can we make this a private symbol and avoid the use of a pointer to prevent memory allocations? See `ExistenceOnly` for an example.

hrpc/get.go
```diff
@@ -18,8 +19,41 @@ type Get struct {
...
+}
+
+// QueryConsistency is used to specify the required consistency of data
+//
+// See https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.2.9/bk_hadoop-ha/content/ha-hbase-timeline-consistency.html
+type QueryConsistency int
+
+const (
+	// Use HBase's default
+	DefaultConsistency QueryConsistency = iota
```
Lekensteyn, line 34: The default is strong consistency, perhaps we could remove this?
```
	Default_Get_Consistency      = Consistency_STRONG
...
func (x *Get) GetConsistency() Consistency {
	if x != nil && x.Consistency != nil {
		return *x.Consistency
	}
	return Default_Get_Consistency
}
```

hrpc/get.go
bwesterb: Made private.

The pointer _here_ is not the issue — the pointer in the protocol message is. There is no analogue of `proto.Bool` which `ExistenceOnly` uses. What we can do is have two global variables for each of the values of `pb.Consistency`. I think it's up to the maintainer if that clutter is worthwhile.

dethi:
Thanks for the PR! I will try to review it before the EOW.

hrpc/get.go
```diff
@@ -156,3 +194,20 @@ func familiesToColumn(families map[string][]string) []*pb.Column {
 	}
 	return cols
 }
+
+// Consistency is a Scan or Get option that requests the given
+// consistency of data.
+func Consistency(consistency QueryConsistency) func(Call) error {
```
dethi, line 200: You can then move this to `query.go` (you will find similar function there)

hrpc/get.go
dethi: Why would we have `DefaultConsistency` when HBase proto doesn't have one and if the user try to use this value, it will just panic. Better not have it at all no?

hrpc/get.go
```diff
@@ -18,8 +19,42 @@ type Get struct {
 	baseQuery
 	// Don't return any KeyValue, just say whether the row key exists in the
 	// table or not.
-	existsOnly bool
-	skipbatch  bool
+	existsOnly  bool
+	skipbatch   bool
+	consistency QueryConsistency
```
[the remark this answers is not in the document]
bwesterb: Ah, missed that. Done.

hrpc/get.go
bwesterb: Done.

hrpc/get.go
```diff
@@ -18,8 +19,42 @@ type Get struct {
...
+}
+
+// QueryConsistency is used to specify the required consistency of data
+//
+// See https://docs.cloudera.com/HDPDocuments/HDP2/HDP-2.2.9/bk_hadoop-ha/
+//          content/ha-hbase-timeline-consistency.html
+type QueryConsistency int
```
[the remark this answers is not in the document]
bwesterb: Done.

dethi:
@bwesterb Thanks for the update. What do you think of my question regarding `DefaultConsistency`? Do you have a use case that would require it?

hrpc/get.go
dethi: > Setting it will not cause a panic
Ah yes sorry I misread that part. Make sense to me. Let's ship it!

dethi (approved):

dethi:
Thanks for your contribution @bwesterb! 😃

bwesterb:
Thanks for the package and the review!

PR #181, opened 2022-02-10
avoid cache avalanche

xpineal:
At the beginning of client start, there is no table region info in cache.
If there are many go routine to access a table at same time, they will miss cache and send request package to get table region info.
In this case, too many duplicated requests should be avoid.

### DIFF
--- a/client.go
+++ b/client.go
@@ -97,6 +97,10 @@ type client struct {
 		string, time.Duration, compression.Codec) hrpc.RegionClient
 
 	compressionCodec compression.Codec
+
+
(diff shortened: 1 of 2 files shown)


dethi:
This seems a little bit similar to https://github.com/tsuna/gohbase/pull/164, not in the implementation but in what it tries to do.

tsuna:
Discussed this with @dethi today and after looking at data on our clusters with @aaronbee we decided to add an option to the client to perform a cache warm up during initialization by preloading all entries from meta for tables of interest.

PR #184, opened 2022-03-15
ci: run test with go1.13 and latest go version

dethi:


tsuna:
Why would anyone still use Go 1.13 at this point?

Issue #174, opened 2021-11-18
RPC objects should not embed a Context


tsuna:
Not only is it a bad practice to store the `Context` in a struct, but it actually led to a race condition being introduced when we added tracing (being fixed by PR #173). We should pass the `Context` along with the RPC, and when we derive a new context due to tracing we just pass along the derived context.

PR #194, opened 2022-08-15
Optimizations and cleanups

aaronbee:
Optimizes some hot spots seen during scale testing. The largest changes are around optimizing `toProto`.

```
name                           old time/op    new time/op    delta
MultiToProto/cellblocks-16       7.44µs ± 1%    5.40µs ± 1%  -27.34%  (p=0.000 n=9+9)
MultiToProto/no_cellblocks-16    5.22µs ± 0%    5.22µs ± 1%     ~     (p=0.423 n=10+10)

name                           old alloc/op   new alloc/op   delta
MultiToProto/cellblocks-16       6.63kB ± 0%    3.42kB ± 0%  -48.40%  (p=0.000 n=10+10)
MultiToProto/no_cellblocks-16    3.84kB ± 0%    3.84kB ± 0%     ~     (all equal)

name                           old allocs/op  new allocs/op  delta
MultiToProto/cellblocks-16         70.0 ± 0%      58.0 ± 0%  -17.14%  (p=0.000 n=10+10)
MultiToProto/no_cellblocks-16      64.0 ± 0%      64.0 ± 0%     ~     (all equal)
```


aaronbee:
I still have some testing to do on a cluster to ensure these changes don't break anything.

region/multi.go
```diff
@@ -25,8 +25,14 @@ var multiPool = sync.Pool{
 }
 
 func freeMulti(m *multi) {
+	// Set pointers to nil to allow the objects to be garbage
```
dethi, line 28: TIL. I originally thought "this can't be true, Go should be smart enough to realise that slice semantic doesn't allow us to access index >= len(slice) and so it should GC them", but no :(

hrpc/mutate.go
```diff
@@ -371,61 +371,40 @@ func (m *Mutate) valuesToProto(ts *uint64) []*pb.MutationProto_ColumnValue {
...
+func toCellblock(row, family, qualifier, value []byte, ts uint64, typ byte) []byte {
 	// keyvaluelength
 	keylength := 2 /* row length */ + len(row) + 1 /* columnfamily length */ +
 		len(family) + len(qualifier) + 8 /* timestamp length */ + 1 /* key type length */
 	valuelength := len(value)
 
 	keyvaluelength := 8 + uint32(keylength+valuelength)
+	cellblock := make([]byte, keyvaluelength+4)
 
-	// keyvaluelength
```
dethi, line 390: Would be nice to keep the comments for clarity

dethi:
You will need to rebase after https://github.com/tsuna/gohbase/pull/195 is merged, that will fix the lint errors.

hrpc/mutate.go
aaronbee: I had felt the code was more clear now so didn't need the comments. I'll take another look.

aaronbee: I added a comment with the the layout of a cellblock. I didn't think the line-by-line comments of the form `copy(cellbock[i:], foo) // adding foo` were very useful. Let me know what you think.

dethi:
Eh the data race is because of the embedded context again https://github.com/tsuna/gohbase/issues/174 😞

aaronbee:
Bummer. I'll see what it would take to fix that.

dethi:
If you call `m.SetContext(nil)` in `freeMulti` that would unblock here. I don't think you should try to fix https://github.com/tsuna/gohbase/issues/174 in this PR. It's a much bigger change (a welcome change but probably best to have it separated)

aaronbee:
Oh, wow, I see the issue now. That is a bit scary. I'll leave out zeroing out the ctx field in freeMulti until #174 is resolved.

dethi:
The change LGTM, but it would be nice to have someone else take a look as well before I merge it.

rodicus (approved):

region/client.go
```diff
@@ -705,20 +700,24 @@ func (c *client) send(rpc hrpc.Call) (uint32, error) {
 	id := c.registerRPC(rpc)
 	header.CallId = &id
 
-	b := newBuffer(4)
-	defer func() { freeBuffer(b) }()
+	headerSize := proto.Size(header)
```
rodicus, line 703: It'd be nice to have comments (similar to the commit's comments) to explain some of the magic calculations going on here

rodicus (changes requested):

region/multi.go
```diff
@@ -25,8 +25,15 @@ var multiPool = sync.Pool{
...
+	// collected
+
+	// m.ctx = nil Do this after #174 is fixed
+	for i := range m.calls {
+		m.calls[i] = nil
+	}
 	m.calls = m.calls[:0]
```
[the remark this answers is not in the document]
aaronbee: I could explain this better in the comments. m.calls is reused, ie. it gets appended to. By not nilling out the slice we can avoid reallocations from appends. m.regions on the other hand is just reassigned to a new slice, so there is no point in keeping the old slice around.

region/client.go
aaronbee: Will do.

region/client.go
```diff
@@ -705,25 +700,59 @@ func (c *client) send(rpc hrpc.Call) (uint32, error) {
...
+	b, err = proto.MarshalOptions{UseCachedSize: true}.MarshalAppend(b, request)
 	if err != nil {
 		return id, fmt.Errorf("failed to marshal request: %s", err)
 	}
-
-	binary.BigEndian.PutUint32(b, uint32(len(b))+cellblocksLen-4)
+	if len(b) != 4+protobufLen {
+		// This shouldn't ever happen, if it does there is a bug in
+		// the size calculation.
+		panic(fmt.Errorf("size of data sent doesn't match expected. expected: %d actual: %d"+
```
rodicus, line 752: nit: did you mean to use fmt.Sprintf instead of fmt.Errorf?

aaronbee: It doesn't actually matter, since panic takes in an interface{} as an argument, but it's common practice to pass an error to a panic. Either way the formatted string will be printed out if it hits.

aaronbee: To actually answer your question, yes I intended to use fmt.Errorf here.

rodicus (approved):
LGTM

dethi (approved):

PR #201, opened 2022-09-23
Gohbase debug state

taran1598:
Motivation: This change adds a new function to the gohbase/client.go to retrieve information about the state of the client so we can debug issues with the client more easily

Change: Added a custom marshaller method to RegionInfo, RegionClient, keyRegionCache, and clientRegionCache to retrieve relevant information. The information is returned in JSON format. An error is returned if the marshaller encountered an error


hrpc/call.go
```diff
@@ -33,6 +34,7 @@ type RegionInfo interface {
 	Table() []byte
 	SetClient(RegionClient)
 	Client() RegionClient
+	json.Marshaler
```
aaronbee, line 37: You don't gain much by adding the json.Marshaler interface to the RegionInfo interface. All you really need is for region.info to implement MarshalJSON. So, I suggest you don't add this here. These interfaces are already way too big.

hrpc/call.go
```diff
@@ -44,6 +46,7 @@ type RegionClient interface {
 	Addr() string
 	QueueRPC(Call)
 	String() string
+	json.Marshaler
```
aaronbee, line 49: Same as above.

caches.go
```diff
@@ -121,6 +149,36 @@ func (krc *keyRegionCache) get(key []byte) ([]byte, hrpc.RegionInfo) {
...
 
+// reads whole b tree in keyRegionCache and gathers debug info
+func (krc *keyRegionCache) KeyRegionCacheDebugInfo(
+	keyRegionInfosMap map[string]hrpc.RegionInfo) map[string]string {
+	regionCacheMap := map[string]string{}
+
+	krc.m.RLock()
+	enum, err := krc.regions.SeekFirst()
+	if err != nil {
+		log.Infof("No regions cached in keyRegionCache: %v", err)
```
rodicus, line 160: No need to log

caches.go
```diff
@@ -92,6 +93,33 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
 	return downregions
 }
 
+// Collects information about the clientRegion cache and stores them in two maps to reduce
+// duplication of data. We do this in one function to avoid running the iterations twice
+func (rcc *clientRegionCache) ClientRegionCacheDebugInfo(
```
rodicus, line 98: just "debugInfo" (unexported)

caches.go
```diff
@@ -92,6 +93,33 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
...
+	keyRegionInfosMap map[string]hrpc.RegionInfo,
```
rodicus, line 99: just "regions"

caches.go
```diff
@@ -92,6 +93,33 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
...
+	clientRegionsMap map[string]hrpc.RegionClient) map[string][]string {
```
rodicus, line 100: nit: "clients"

caches.go
```diff
@@ -92,6 +93,33 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
...
+	// key = RegionClient memory address , value = List of RegionInfo addresses
+	clientRegionCacheMap := map[string][]string{}
+
+	rcc.m.RLock()
+	for client, reginfos := range rcc.regions {
+		clientRegionInfoMap := make([]string, 0)
```
rodicus, line 106: Pass len(reginfos) as the second arg to "make". You could then do the following on line 112:

  clientRegionInfoMap[i] = fmt.Sprintf(..)

instead of using "append"

debug_state_test.go
```diff
@@ -0,0 +1,94 @@
...
+	region3.SetClient(regClient)
+	client.clients.put("regionserver:1", region3, newClientFn)
+
+	json, err := DebugState(client)
+
+	if err != nil {
+		t.Errorf("DebugInfo should not have an error: %v", err)
+	}
+
+	log.Println(string(json))
```
rodicus, line 92: Not needed. Instead validate the json output with json.Validate() or something like that

integration_test.go
```diff
@@ -2275,3 +2278,47 @@ func TestMoveRegion(t *testing.T) {
...
+	val := []byte("1")
+	if host == nil {
+		t.Fatal("Host is not set!")
+	}
+
+	c := gohbase.NewClient(*host)
+	defer c.Close()
+	err := insertKeyValue(c, key, "cf", val)
+	if err != nil {
+		t.Errorf("Put returned an error: %v", err)
```
rodicus, line 2293: Change to Fatal

integration_test.go
```diff
@@ -2275,3 +2278,47 @@ func TestMoveRegion(t *testing.T) {
...
+	}
+
+	jsonVal, err := gohbase.DebugState(c)
+
+	if err != nil {
+		t.Errorf("DebugState returned an error when it shouldn't have: %v", err)
+	}
+
+	fmt.Println(string(jsonVal))
```
rodicus, line 2302: Use t.Logf or remove

client.go
```diff
@@ -27,6 +28,18 @@ const (
...
+	ClientTypeJsonKey          = "ClientType"
+	ClientRegionMapJsonKey     = "ClientRegionMap"
+	RegionInfoMapJsonKey       = "RegionInfoMap"
+	KeyRegionCacheJson         = "KeyRegionCache"
+	ClientRegionCacheJsonKey   = "ClientRegionCache"
+	MetaRegionInfoJsonKey      = "MetaRegionInfo"
+	AdminRegionInfoJsonKey     = "AdminRegionInfo"
+	Done_StatusJsonKey         = "Done_Status"
+	RegionLookupTimeoutJsonKey = "RegionLookupTimeout"
+	RegionReadTimeoutJsonKey   = "RegionReadTimeout"
```
rodicus, line 42: Remove them. Just use the plain string in the test if we're checking for those json elements

caches.go
```diff
@@ -92,6 +93,36 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
 	return downregions
 }
 
+// Collects information about the clientRegion cache and stores them in two maps to reduce
+// duplication of data. We do this in one function to avoid running the iterations twice
```
rodicus, line 97: Instead of "store them in two maps" mention that it appends the data into the given maps

caches.go
```diff
@@ -121,6 +152,34 @@ func (krc *keyRegionCache) get(key []byte) ([]byte, hrpc.RegionInfo) {
 	return k.([]byte), v.(hrpc.RegionInfo)
 }
 
+// reads whole b tree in keyRegionCache and gathers debug info
```
rodicus, line 155: Maybe mention that this appends entries to the given map

caches.go
```diff
@@ -92,6 +93,36 @@ func (rcc *clientRegionCache) clientDown(c hrpc.RegionClient) map[hrpc.RegionInf
 	return downregions
 }
 
+// Collects information about the clientRegion cache and stores them in two maps to reduce
```
rodicus, line 96: nit: the code could look more organized if all the new functions appear together at the end of the file

client.go
```diff
@@ -142,6 +143,70 @@ func newClient(zkquorum string, options ...Option) *client {
...
+func DebugState(client Client) ([]byte, error) {
+
+	debugInfoJson, err := json.Marshal(client)
+	if err != nil {
+		log.Errorf("Cannot turn client into JSON bytes array: %v", err)
+	}
+	return debugInfoJson, err
+}
+
+func (c *client) DoneStatus() string {
```
rodicus, line 156: No need to export (capitalize) this function. Since it's only used by MarshalJSON this code could just be moved in that function instead

debug_state_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/tsuna/gohbase/hrpc"
+	"github.com/tsuna/gohbase/region"
+)
+
+func TestDebugStateSanity(t *testing.T) {
+	client := newClient("~invalid.quorum~") // We shouldn't connect to ZK.
+
+	reg := client.getRegionFromCache([]byte("test"), []byte("theKey"))
+	if reg != nil {
```
rodicus, line 17: Remove. Not relevant since we're not testing an initial empty cache

region/client.go
```diff
@@ -265,6 +266,28 @@ func (c *client) inFlightDown() error {
 	return nil
 }
 
+// InFlight returns inFlight value
```
rodicus, line 269: Move extra functions to MarshalJSON since they're only needed by it

rodicus: Thanks for moving these functionality, but the original functions still appear on the review

rodicus (approved):
LGTM

dethi (approved):

PR #202, opened 2022-10-06
Add SendBatch API

### LINKED ISSUE #68: Manually build multi-actions?
It looks like this library automatically bundles up multiple simultaneous requests into a single multi-action. Is it possible for a consumer of the library to do batch requests easier than making multiple requests in many different goroutines?

### LINKED ISSUE #117: gohbase support batch gets now? if yes, how to use pls

aaronbee:
Added the API:
```
SendBatch(ctx context.Context, batch []hrpc.Call) (res []hrpc.RPCResult, allOK bool)
```
The res has results in the same order as the RPCs in batch. Each RPCResult contains a result and an error. If the error is non-nil then the RPC failed or was not attempted. If it's nil the batch succeeded. The allOK is false if any of the results contain an error.

closes #68, closes #117

### DIFF
--- a/region/multi.go
+++ b/region/multi.go
@@ -279,9 +279,9 @@ func (m *multi) returnResults(msg proto.Message, err error) {
 }
 
 // add adds the call and returns wether the batch is full.
-func (m *multi) add(call hrpc.Call) bool {
-	m.calls = append(m.calls, call)
-	return len(m.calls) == m.size
+func (m *multi) add(calls []hrpc.Call) bool {
+	m.calls = append(m.calls, calls...)
+	return len(m.calls) >= m.size
 }
 
 // len returns number of batched calls.
--- a/rpc.go
+++ b/rpc.go
@@ -250,14 +250,9 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
 	)
 	wg.Add(len(rpcByClient))
 	for client, rpcs := range rpcByClient {
-		// TODO: Move this to the RegionClient interface so we don't
-		// need to type assert here
-		qb := client.(interface {
-			QueueBatch(ctx context.Context, rpcs []hrpc.Call)
-		})
 		go func(client hrpc.RegionClient, rpcs []hrpc.Call) {
 			defer wg.Done()
-			qb.QueueBatch(ctx, rpcs)
+			client.QueueBatch(ctx, rpcs)
 			ctx, sp := observability.StartSpan(ctx, "waitForResult")
 			defer sp.End()
 			ok := c.waitForCompletion(ctx, client, rpcs, res, rpcToRes)
--- a/rpc.go
+++ b/rpc.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"math"
 	"strconv"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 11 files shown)


region/multi.go
```diff
@@ -273,9 +273,9 @@ func (m *multi) returnResults(msg proto.Message, err error) {
 }
 
 // add adds the call and returns wether the batch is full.
```
ciacono, line 275: typo

ciacono (commented):
Initial pass

rpc.go
```diff
@@ -68,6 +69,89 @@ func (c *client) getRegionForRpc(ctx context.Context, rpc hrpc.Call) (hrpc.Regio
 	return nil, ErrCannotFindRegion
 }
 
+// SendBatch will execute all the calls in the Batch. Batches should
+// not be reused after passing to SendBatch.
```
ciacono, line 73: Do clients directly call SendBatch or is it used by other methods in the API? Should we ensure a batch cannot be reused vs saying not to?
If so could add a check in` func (b *Batch) Add(m *hrpc.Mutate)` in batch.go too

aaronbee: Clients will call SendBatch directly. I guess there really isn't any harm to a client reusing a batch, so I could actually remove this note. At the point that it does become an issue, because the batch keeps some state we could panic in that scenario.

batch.go
```diff
@@ -0,0 +1,51 @@
...
+// Do we need this?
+func (b *Batch) Add(m *hrpc.Mutate) {
+	b.checkCall(m)
+	b.ms = append(b.ms, m)
+}
+
+func (b *Batch) checkCall(m *hrpc.Mutate) {
+	table := m.Table()
+	if !bytes.Equal(table, b.table) {
+		panic("Call's table should the same as the batch's table")
```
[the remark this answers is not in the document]
aaronbee: I used a panic here because if the user gets this wrong, it's likely a programming bug rather than an error that can be handled.

aaronbee:
Latest version changes the API of SendBatch and gets rid of the Batch type:

New API:

`SendBatch(ctx context.Context, batch []hrpc.Call) (res []hrpc.RPCResult, err error)`

The `res` has results in the same order as the RPCs in batch. Each RPCResult contains a result and an error. If the error is non-nil then the RPC failed or was not attempted. If it's nil the batch succeeded. The error return is just set to non-nil if any of the RPCResults contain an error.

region/multi.go
aaronbee: Deleted

batch.go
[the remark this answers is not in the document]
aaronbee: In the latest version this is now an error.

batch.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	if m.SkipBatch() {
+		panic("Call is not batchable")
```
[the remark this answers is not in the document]
aaronbee: This is now an error.

aaronbee:
Progress update: The new API feels better than my first attempt. This change is still missing tests.

aaronbee:
Added unit tests for SendBatch. I think there are just a couple small changes I want to make. The change is ready for review.

dethi:
Closes #68 #117

rpc.go
```diff
@@ -158,6 +159,182 @@ func (c *client) getRegionAndClientForRPC(ctx context.Context, rpc hrpc.Call) (
...
+
+	// Send each group of RPCs to region client to be executed.
+	var (
+		wg sync.WaitGroup
+
+		mu   sync.Mutex
+		fail bool
+	)
+	wg.Add(len(rpcByClient))
+	for client, rpcs := range rpcByClient {
```
ciacono, line 252: It looks like in `findClients`, if there's any error (ex. missing regions), res will be updated for those calls and that will be a batch error. It seems to me in that case that there would not be a need to send those rpcs to `QueueBatch` and `waitForCompletion` since they've already been errored out? Could that save some time/work?

rpc.go
```diff
@@ -158,6 +159,182 @@ func (c *client) getRegionAndClientForRPC(ctx context.Context, rpc hrpc.Call) (
...
+// the same table and must be Batchable.
+//
+// SendBatch will discover the correct region and region server for
+// each Call and dispatch the Calls accordingly. SendBatch is not an
+// atomic operation. Some calls may fail and others succeed. Calls
+// sharing a region will execute in the order passed into SendBatch.
+//
+// SendBatch returns a slice of [hrpc.RPCResult] each containing a
+// response and an error. The results will be returned in the same
+// order as the Calls in the batch, in other wrods the i'th result
```
ciacono, line 179: Nit: typo

rpc.go
aaronbee: If there is any error in `findClient` then the whole `SendBatch` is aborted, it returns on line 241. The RPCs with missing regions will get the missing region error and all the other ones will have `NotExecutedError`.

rpc.go
aaronbee: Fixed.

rpc.go
ciacono: > all the other ones will have NotExecutedError

This makes sense and sounds better than trying with the RPCs that do have clients.

Thanks!

rpc.go
```diff
@@ -158,6 +159,182 @@ func (c *client) getRegionAndClientForRPC(ctx context.Context, rpc hrpc.Call) (
...
+		} else if b, batchable := rpc.(hrpc.Batchable); !batchable || b.SkipBatch() {
+			res[i].Error = errors.New("non-batchable call passed to SendBatch")
+			allOK = false
+		}
+	}
+	if !allOK {
+		return res, allOK
+	}
+
+	rpcByClient, ok := c.findClients(ctx, batch, res)
```
dethi, line 239: Could interesting to add an histogram to track how many regions are "targeted" by batch of RPC. To be added in following PR maybe.

dethi (approved):
Overall, LGTM Aaron

Issue #210, opened 2023-02-08
OpenTelemetry can be set as optional


ngoctd314:
gohbase auto use opentelmetry, but i don't want to send any trace about gohbase to opentelemetry. I think it must be optional.

PR #211, opened 2023-02-15
openTelemetry as optional

ngoctd314:
i have a basic solution for this issue
https://github.com/tsuna/gohbase/issues/210

### DIFF
--- a/internal/observability/noop_span.go
+++ b/internal/observability/noop_span.go
@@ -1,0 +1,77 @@
+package observability
+
+import (
+	"sync"
+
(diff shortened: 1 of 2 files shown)


internal/observability/noop_span.go
```diff
@@ -0,0 +1,71 @@
...
+	// noop
+	return trace.NewNoopTracerProvider()
+}
+
+func newNoopSpan() trace.Span {
+	return noopTraceSpan{}
+}
+
+var (
+	isNoopTracing = false
```
dethi, line 64: I don't think using a global variable here is the right approach.

PR #215, opened 2023-03-21
region/client: send request timeout to HBase

dethi:
TODO:
- how to write a test for this?
- what exception is returned?
- are we guaranteed to get a reply if the request timeout?
- what if the requests is dropped from the queue?


region/client.go
```diff
@@ -626,6 +626,15 @@ func (c *client) send(rpc hrpc.Call) (uint32, error) {
 		RequestParam: proto.Bool(true),
 	}
 
+	deadline, ok := rpc.Context().Deadline()
```
aaronbee, line 629: This isn't going to work for requests sent in multi's, ie. all mutate requests, since a multi's Context() method just returns context.Background().

Maybe you don't care about that case. Just wanted to raise it in case you weren't aware. If you did want it to work for multi's you'd probably have to find the latest Deadline among the requests and use that as your timeout.

Issue #186, opened 2022-05-03
gohbase stuck retrying RPCs after region server temporarily down


rodicus:
We are experiencing an issue where Put RPCs (and possibly any other types of RPCs) for a specific region may end up continuously failing with a `ServerError`.
This seems to be triggered by HBase moving regions after one of the region servers goes down temporarily.
The current workaround is to restart the gohbase client process, which indicates that the issue is with gohbase and not HBase itself.

Our guess is that a region client connection is closed, but it remains associated with one or more region info instances in the gohbase cache.
Probably after a disconnection the region client is not properly cleaned up and may end up being reused in RPC operations.

We have seen the issue on a production HBase cluster that was periodically having sync failures making region servers restart multiple times a day.
We have tried reproducing this issue in a lab by bringing region servers down temporarily, but the cluster and client process always recover.

The gohbase versions we have seen this failed RPCs include:
6acaa09d18ed
348991136365
which are relatively old, but there doesn’t appear to be any changes newer in the repo that could have potentially resolved the issue in question.

Below is a breakdown of goroutines of version 348991136365 while RPCs are failing. In this setup there are 3 HBase region servers and 8000 workers issuing `Put` RPCs:
```
3 goroutines in the "send" loop (one per server?)
/go/src/arista/vendor/github.com/tsuna/gohbase/region/new.go:69 +0x c.Dial go c.processRPCs()
/go/src/arista/vendor/github.com/tsuna/gohbase/region/client.go:364 +0x c.processRPCs select // wait for the next batch

3 goroutines in the "receive" loop (one per server?)
/go/src/arista/vendor/github.com/tsuna/gohbase/region/new.go:71 +0x c.Dial go c.receiveRPCs()
/go/src/arista/vendor/github.com/tsuna/gohbase/region/client.go:444 +0x c.receiveRPCs c.receive()
/go/src/arista/vendor/github.com/tsuna/gohbase/region/client.go:464 +0x c.receive c.readFully(sz[:])
/go/src/arista/vendor/github.com/tsuna/gohbase/region/client.go:580 +0x c.readFully io.ReadFull

3 "Put" goroutines queueing an RPC
/go/src/arista/vendor/github.com/tsuna/gohbase/client.go:237 +0x c.Put c.mutate(p)
/go/src/arista/vendor/github.com/tsuna/gohbase/client.go:264 +0x c.mutate c.SendRPC(m)
[...]

PR #227, opened 2023-07-29
Simplify and document region life cycle

aaronbee:
* Removed RegionInfo.SetClient. The client is now set to nil by MarkUnavailable, and then set again with MarkAvailable. This removes bugs and races where a region info may have its Client set when it shouldn't be. A region now has 3 distinct states it may be in:

1. Unavailable: AvailabilityChan() is non-nil and Client() is nil
2. Available: AvailabilityChan() is nil and Client() is non-nil
3. Stale: AvailabilityChan() is nil and Client() is nil

* Documented RegionInfo type with information about those 3 states and the lifecycle.

* Documented establishRegion with what it does and what it's requirements are.

* Removed calls to SetClient(nil)/MarkUnavailable in clientRegionCache methods. That work is the responsibility of the caller.

* getRegionAndClientForRPC no longer marks a region as unavailable if Client is nil after waiting on the AvailabilityChan. If Client is nil, that indicates the region is stale and a lookup has to be performed. Marking it as unavailable and trying to reestablish it is wasted effort.

* The above change broke the use of the meta and admin regions, which are created when creating the gohbase Client/AdminClient, but don't have a Client set. These regions were relying on a missing Client to cause the region to be established. Now the meta and admin regions are established as part of NewClient and NewAdminClient, respectively.

closes: #186


aaronbee:
This change still needs some work.

PR #229, opened 2023-08-15
region: Use a buffered reader

aaronbee:
Wrap connection in a bufio.Reader for reading RPC responses from RegionServer. This should reduce the number of read system calls by attempting to fill a 4KB buffer on each read.

Some mock tests that expected certain sizes of buffers passed to Read() had to be made more lenient.

### DIFF
--- a/region/client.go
+++ b/region/client.go
@@ -6,6 +6,7 @@
 package region
 
 import (
+	"bufio"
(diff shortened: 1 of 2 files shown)


dethi (approved):
Nice find!

PR #230, opened 2023-08-15
SendBatch: Launch one fewer goroutine

aaronbee:
A small optimization to reduce the number of goroutines launched by SendBatch, while achieving the same parallelism. This is especially nice for SendBatch calls that only need to reach one Region Server, because then 0 additional goroutines need to be launched.

Also, changed `fail` from a bool + mutex to a uint32 that is atomically set to 1 on failure. This was primarily done to reduce lines of code.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -14,6 +14,7 @@ import (
 	"math"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


rpc.go
```diff
@@ -243,30 +244,36 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
...
+		if i++; i < clientCount {
+			go sendBlocking(client, rpcs)
+		} else {
+			// Small optimization: don't launch a goroutine for the
+			// last client
+			sendBlocking(client, rpcs)
+		}
 	}
 	wg.Wait()
-	allOK = !fail
```
dethi, line 267: This line need to be kept, otherwise the defer will not report correctly if the batch failed or not in the metric / trace. (see top of the function)

dethi (changes requested):

rpc.go
aaronbee: Good catch.

aaronbee:
I created https://github.com/tsuna/gohbase/pull/232 as an alternative to this PR.

aaronbee:
Superseded by #232

PR #232, opened 2023-08-24
SendBatch: Don't start goroutines

aaronbee:
Instead of starting goroutines we can queue each region server's collection of RPCs and only after queueing them all wait for all responses. We get a similar amount of parallelism, the only potential slow down is that we can't queue one region server's RPCs until the previous is able to be queued.

By not starting goroutines, the resource usage of SendBatch is more predictable. And not starting goroutines is likely to be less expensive overall.

This change also adds a new metric for number of splits that occur in SendBatch.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -239,6 +239,7 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
 	if !ok {
 		return res, false
 	}
+	sendBatchSplitCount.Observe(float64(len(rpcByClient)))
 
 	// Send each group of RPCs to region client to be executed.
 	type clientAndRPCs struct {
--- a/rpc.go
+++ b/rpc.go
@@ -13,7 +13,6 @@ import (
 	"io"
 	"math"
 	"strconv"
-	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


aaronbee:
Testing shows this didn't affect performance, so I think this is worth merging because it simplifies thinking about the cost of a SendBatch.

rpc.go
```diff
@@ -240,30 +239,30 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
...
-				mu.Lock()
-				fail = true
-				mu.Unlock()
-			}
-		}(client, rpcs)
+		client.QueueBatch(ctx, rpcs)
+		cAndRs = append(cAndRs, clientAndRPCs{client, rpcs})
+	}
+
+	ctx, sp2 := observability.StartSpan(ctx, "waitForResult")
```
ciacono, line 257: Should this also be named something different (like `sp2`) because there's the initial `ctx` and `sp` in the deferred function at the beginning and could capture this `ctx` instead?

ciacono (commented):
The idea behind this makes sense to me, although I do see concern with the potential slowdown because we are doing this sequentially. But having the resource usage being more predictable (and less expensive with less goroutines) seems worthwhile. Other than the comment I left, LGTM.

rpc.go
aaronbee: Good catch. Jeez, I hate that this tracing stuff makes me have to think about the correct context to use...

I can move this code into it's own function to simplify things.

aaronbee: Fixed.

ciacono (approved):

PR #234, opened 2023-09-04
hrpc: allow table creation with table attributes

### LINKED ISSUE #161: CreateTable should allow table attributes
Currently we can only pass [ColumnFamilySchema attributes](https://github.com/tsuna/gohbase/blob/master/pb/HBase.proto#L54)  in `CreateTable`. Need to be able to specify [TableShema attributes](https://github.com/tsuna/gohbase/blob/master/pb/HBase.proto#L43)

dethi:
Closes #161

### DIFF
--- a/table_test.go
+++ b/table_test.go
@@ -30,8 +30,10 @@ var infoFamily = map[string][]string{
 }
 
 var cFamilies = map[string]map[string]string{
-	"cf":  nil,
-	"cf2": nil,
+	"cf": nil,
+	"cf2": {
+		"MIN_VERSIONS": "1",
+	},
 }
 
 func TestCreateTable(t *testing.T) {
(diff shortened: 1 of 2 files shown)


table_test.go
```diff
@@ -114,6 +116,52 @@ func TestCreatePresplitTable(t *testing.T) {
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+		rsp = append(rsp, res)
+	}
+	if len(rsp) != 1 {
+		t.Errorf("Meta returned %d rows for prefix '%s' , want 1", len(rsp), metaKey)
```
aaronbee, line 161: This is checking that the table was created successfully. Are we able to check that the desired attribute is set?

dethi: Right, the test isn’t great. Gohbase itself doesn’t have a method to get the attribute back afaik. Could potentially be added

aaronbee: Ack

aaronbee (approved):

PR #239, opened 2023-11-04
region: Reduce allocations of RequestHeader in send

aaronbee:
RequestHeaders are created for every request. Put them in a sync.Pool to reduce allocations.


region/client.go
```diff
@@ -701,6 +715,7 @@ func (c *client) send(rpc hrpc.Call) (uint32, error) {
 	if err != nil {
 		return id, fmt.Errorf("failed to marshal request header: %s", err)
 	}
+	dropRequestHeader(header)
```
ciacono, line 718: Why `dropRequestHeader` here instead of deferring it to the end after `header := getRequestHeader`?
Wouldn't calling `dropRequestHeader` here interfere with the panic error message below (if it does ever happen, even though it shouldn't)?

aaronbee: I want to return the header object for reuse as soon as possible, to make it more likely it will be reused/there will be fewer allocated at one time.
This is the earliest line where it can be dropped. If I used a defer we would continue to hold onto the header object through the point where we do the write on the socket, which may take a while to complete. The fact that we don't return it on error cases is not a big deal, it just means the object will be GC'd rather than reused.

That all said, I would like to pull out this code that builds up the message into its own function that is separate from the write on the socket. With that done I could use a defer to release the header object in a more reasonable place. And with that code pulled out into its own function we can write better tests and benchmarks for it.

dethi:
Maybe we can do what we discussed previously, introduce vtprotobuf and use their pooling mechanism? May be easier than adding this is different place manually

aaronbee:
> Maybe we can introduce vtprotobuf and use their pooling mechanism (instead of protobuf-go)? May be easier than adding this is different place manually

Yeah, that's a better idea. This was just a quick and easy change, but doesn't actually address some of the bigger allocations we see that come from some of the other protobuf objects, like `ScanRequest`. I can investigate the vrprotobuf pool code to see if that makes it easier to address things like the `ScanRequest`, which is allocated in one package, and used in another.

ciacono (approved):
Well, this LGTM. I think you could merge this in and look into `vtprotobuf` going forward?

dethi (approved):

region/client.go
aaronbee: Refactored the code.

aaronbee:
> > Maybe we can introduce vtprotobuf and use their pooling mechanism (instead of protobuf-go)? May be easier than adding this is different place manually
> 
> Yeah, that's a better idea. This was just a quick and easy change, but doesn't actually address some of the bigger allocations we see that come from some of the other protobuf objects, like `ScanRequest`. I can investigate the vrprotobuf pool code to see if that makes it easier to address things like the `ScanRequest`, which is allocated in one package, and used in another.

I have started playing with vtprotobuf. Some findings:
* My new `marshalProto` func is > 2x faster when using the VT protobuf methods.
* The pool code it generates is not rocket science. Basically just creates a pool for a certain protobuf type, almost exactly like the code I have in this PR. There are cases where I want to do something more than just this basic thing, so I may not rely on the vtproto generated code and just write it myself. For example a ScanRequest contains a Scan object. We always create both, so when we pool we should pool them together.

I will post another PR for these changes.

PR #242, opened 2023-11-16
cache: add a metric for the region cache size

dethi:


caches.go
```diff
@@ -319,6 +320,7 @@ func (krc *keyRegionCache) del(reg hrpc.RegionInfo) bool {
 	// let region establishers know that they can give up
 	reg.MarkDead()
 
+	cachedRegionTotal.Dec()
```
aaronbee, line 323: You should only decrement here if success is true.

caches.go
```diff
@@ -304,6 +303,8 @@ func (krc *keyRegionCache) put(reg hrpc.RegionInfo) (overlaps []hrpc.RegionInfo,
 	}
 	krc.m.Unlock()
 
+	// -(N overlapped regions) + (1 new region)
+	cachedRegionTotal.Add(-float64(len(overlaps)) + 1)
```
aaronbee, line 307: I think what you have here is correct, but it's a little tricky. Something I would have more confidence in is if you were to get the length before and after modification and add the diff to the counter. ie.

```
krc.m.Lock()
beforeLen := krc.regions.Len()
krc.regions.Put(...
...
for _, o := range overlaps {
    ...
}
afterLen := krc.regions.Len()
krc.m.Unlock()

cachedRegionTotal.Add(float64(afterLen-beforeLen))
```
Up to you, whichever you prefer.

dethi: Good suggestion!

caches.go
dethi: Nice catch!

aaronbee (approved):

PR #249, opened 2024-04-25
Expose Dialers inside Zk and Region

### LINKED ISSUE #248: Expose Dialers for Zookeeper and Region package to allow using proxies
We use your client in an environment where we need to go through a SOCKS5 proxy. It's pretty easy to support this by exposing the dialers within the zookeeper client and your own Region client. I'm filing this issue so that can attach a pull request to it shortly.

mromaszewicz:
Fixes #248

This change allows for overriding the default dialers used by the ZooKeeper client and the Region client to connect to their respective servers. Proxy-aware dialers can be installed by people who need them.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,14 +15,14 @@ import (
 
 	gzk "github.com/go-zookeeper/zk"
 	log "github.com/sirupsen/logrus"
-	"google.golang.org/protobuf/proto"
-	"modernc.org/b/v2"
 
 	"github.com/tsuna/gohbase/compression"
 	"github.com/tsuna/gohbase/hrpc"
 	"github.com/tsuna/gohbase/pb"
 	"github.com/tsuna/gohbase/region"
 	"github.com/tsuna/gohbase/zk"
+	"google.golang.org/protobuf/proto"
+	"modernc.org/b/v2"
 )
 
 const (
--- a/client.go
+++ b/client.go
@@ -13,14 +13,16 @@ import (
 	"sync"
 	"time"
 
+	gzk "github.com/go-zookeeper/zk"
 	log "github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/proto"
+	"modernc.org/b/v2"
+
 	"github.com/tsuna/gohbase/compression"
 	"github.com/tsuna/gohbase/hrpc"
 	"github.com/tsuna/gohbase/pb"
 	"github.com/tsuna/gohbase/region"
 	"github.com/tsuna/gohbase/zk"
-	"google.golang.org/protobuf/proto"
-	"modernc.org/b/v2"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


client.go
```diff
@@ -268,6 +276,23 @@ func CompressionCodec(codec string) Option {
 	}
 }
 
+// ZooKeeperDialer will return an option to pass the given dialer function
+// into the ZooKeeper client Connect() call, which allows for customizing
+// network connections.
+func ZooKeeperDialer(dialer gzk.Dialer) Option {
```
aaronbee, line 282: I'd rather not make a type from github.com/go-zookeeper/zk an explicit part of the API of gohbase. If github.com/go-zookeeper/zk changes the Dialer type or adds a ContextDialer for example then we would have to change gohbase as well. Or gohbase decides to use a different zookeeper library in the future then this API will have to change.

What I would prefer is we make the region Dialer and the zookeeper Dialer the same type. And I think it makes more sense to make it a function rather than an interface (this is consistent with the zookeeper library and with gRPC).

The function should be: `func(ctx context.Context, network, addr string) (net.Conn, error)`. This works easy enough for the region dialer. For the zookeeper dialer you have to do a little work to convert the context dialer func to the zk.Dialer signature:

```go
func makeZKDialer(ctxDialer func(ctx context.Context, network, addr string) (net.Conn, error)) zk.Dialer {
    return func(network, addr string, timeout time.Duration) (net.Conn, error) {
        ctx, cancel := context.WithTimeout(context.Background(), timeout)
        defer cancel()
        return ctxDialer(ctx, network, addr)
    }
}
```

client.go
```diff
@@ -268,6 +276,23 @@ func CompressionCodec(codec string) Option {
...
+	return func(c *client) {
+		c.zkDialer = dialer
+	}
+}
+
+// RegionDialer will return an option that uses the specified Dialer for
+// connecting to region servers. This allows for connecting through proxies.
+func RegionDialer(dialer region.Dialer) Option {
```
aaronbee, line 290: As suggested above, this should take in a `func(ctx context.Context, network, addr string) (net.Conn, error)`.
I don't think we have to give a name to the type.

aaronbee (changes requested):
Thanks for your contribution. This looks good, just a couple comments below on changing the API.

client.go
mromaszewicz: will do.

mromaszewicz:
Requested changes made.

aaronbee (approved):
Thanks!

aaronbee:
We have a linter to prevent long lines:
```
./client.go:101: Line too long (134/100)
./client.go:281: Line too long (103/100)
./rpc_test.go:305: Line too long (144/100)
./zk/client.go:66: Line too long (132/100)
./zk/client.go:129: Line too long (106/100)
./mockrc_test.go:181: Line too long (149/100)
```

mromaszewicz:
> We have a linter to prevent long lines:
> 
> ```
> ./client.go:101: Line too long (134/100)
> ./client.go:281: Line too long (103/100)
> ./rpc_test.go:305: Line too long (144/100)
> ./zk/client.go:66: Line too long (132/100)
> ./zk/client.go:129: Line too long (106/100)
> ./mockrc_test.go:181: Line too long (149/100)
> ```

Yep, i saw the failures and fixed them right away. That anonymous function is very long :)

PR #250, opened 2024-04-30
cache: added region cache warmup

gogochickenleg:
This change adds an option in gohbase to prepopulate its region cache. It does a single scan of the meta table and cache all regions.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2374,3 +2374,41 @@ func TestDebugState(t *testing.T) {
 	assert.Equal(t, 1, len(keyRegionCache.(map[string]interface{})))
 	assert.Equal(t, 1, len(clientRegionCache.(map[string]interface{}))) // only have one client
 }
+
+type regionInfoAndAddr struct {
+	regionInfo hrpc.RegionInfo
+	addr       string
+}
+
+// Test loading region cache
+func TestCacheRegions(t *testing.T) {
+	c := gohbase.NewClient(*host)
+	defer c.Close()
+
+	// make sure region cache is empty at startup
+	var jsonUnMarshalStart map[string]interface{}
+	jsonVal, err := gohbase.DebugState(c)
+	err = json.Unmarshal(jsonVal, &jsonUnMarshalStart)
+	if err != nil {
+		t.Fatalf("Encoutered eror when Unmarshalling: %v", err)
+	}
+	cacheLength := len(jsonUnMarshalStart["KeyRegionCache"].(map[string]interface{}))
+	if cacheLength != 0 {
+		t.Fatal("expected empty region cache when creating a new client")
+	}
+
+	c.CacheRegions([]byte(table))
+
+	var jsonUnMarshalCached map[string]interface{}
+	jsonVal, err = gohbase.DebugState(c)
+	err = json.Unmarshal(jsonVal, &jsonUnMarshalCached)
+	if err != nil {
+		t.Fatalf("Encoutered eror when Unmarshalling: %v", err)
+	}
+	// CreateTable init function starts hbase with 4 regions
+	cacheLength = len(jsonUnMarshalCached["KeyRegionCache"].(map[string]interface{}))
+	if cacheLength != 4 {
+		t.Fatalf("Expect 4 regions but got: %v", cacheLength)
+	}
+
+}
--- a/rpc.go
+++ b/rpc.go
@@ -24,6 +24,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+type regionInfoAndAddr struct {
+	regionInfo hrpc.RegionInfo
+	addr       string
+}
+
 // Constants
 var (
 	// Name of the meta region.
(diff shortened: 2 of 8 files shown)


integration_test.go
```diff
@@ -2374,3 +2374,45 @@ func TestDebugState(t *testing.T) {
...
 	assert.Equal(t, 1, len(clientRegionCache.(map[string]interface{}))) // only have one client
 }
+
+type regionInfoAndAddr struct {
+	regionInfo hrpc.RegionInfo
+	addr       string
+}
+
+// Test looking up meta table for all regions.
+func TestMetaLookupForTable(t *testing.T) {
```
aaronbee, line 2384: I feel like it would be better if this test just called `CacheRegions` on the client and then somehow verified that all the regions were loaded.

rpc.go
```diff
@@ -554,6 +559,102 @@ func (c *client) lookupRegion(ctx context.Context,
...
+			reg = c.adminRegionInfo
+			regs = append(regs, regionInfoAndAddr{regionInfo: reg, addr: addr})
+		} else if bytes.Equal(table, metaTableName) {
+			log.WithField("resource", zk.Meta).Debug("looking up region server of hbase:meta")
+
+			addr, err = c.zkLookup(lookupCtx, zk.Meta)
+			cancel()
+			reg = c.metaRegionInfo
+			regs = append(regs, regionInfoAndAddr{regionInfo: reg, addr: addr})
+		} else {
```
aaronbee, line 586: This code shouldn't be needed I think. You wouldn't use lookupAllRegions to look up the admin or meta regions.

aaronbee (commented):
Thanks for the PR! I left a couple comments. Let me know if they make sense.

rpc.go
```diff
@@ -678,6 +762,52 @@ func (c *client) metaLookup(ctx context.Context,
 	return reg, addr, nil
 }
 
+// Creates the META key to search for all regions
+func CreateAllRegionSearchKey(table []byte) []byte {
```
aaronbee, line 766: This doesn't need to be exported if the integration test is changed.

integration_test.go
aaronbee: This is a bit round-about, but you can call `gohbase.DebugState(c)` to get state about the cache as JSON. You could then parse that json back into maps/structs and check that way.
If that is too cumbersome, I think you could expose a method for getting the cache state.

gogochickenleg: Thanks! I was able to use `DebugState` to check the region cache. I changed the test to check the region cache before and after running `CacheRegions`

aaronbee:
@dethi Do you want to take a final look at this?

rpc.go
```diff
@@ -554,6 +559,85 @@ func (c *client) lookupRegion(ctx context.Context,
...
+			// the cache while we were looking it up.
+			overlaps, replaced := c.regions.put(reg)
+			if !replaced {
+				// the same or younger regions are already in cache
+				continue
+			}
+
+			// otherwise, new region in cache, delete overlaps from client's cache
+			for _, r := range overlaps {
+				c.clients.del(r)
```
dethi, line 630: I will need to look again, I can remember if deleting the region from the cache is enough or if there is other cleanup to do, like marking it as unavailable.

rpc.go
```diff
@@ -554,6 +559,85 @@ func (c *client) lookupRegion(ctx context.Context,
...
+		}
+	}
+}
+
+func (c *client) findAllRegions(ctx context.Context, table []byte) ([]regionInfoAndAddr, error) {
+	regs, err := c.lookupAllRegions(ctx, table)
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < len(regs); i++ {
```
dethi, line 615: Any reason to not just do `for regaddr := range regs { [...] }`

rpc.go
```diff
@@ -554,6 +559,85 @@ func (c *client) lookupRegion(ctx context.Context,
...
+			}
+
+			// otherwise, new region in cache, delete overlaps from client's cache
+			for _, r := range overlaps {
+				c.clients.del(r)
+			}
+		}
+
+		// Start a goroutine to connect to the region
+		go c.establishRegion(reg, addr)
```
dethi, line 635: I'm not sure if we should establish a connection immediately. I would think that just loading the cache is enough and once one of the region is being used, then a connection will be established. No?

dethi:
Left a few comment + CI is complaining about the mock missing the new function.

rpc.go
gogochickenleg: I think in other place in the code where we are deleting the overlapped regions in cache, we don't mark them as unavailable. But maybe @aaronbee can confirm

rpc.go
gogochickenleg: In the `SendRPC` function, we first call `getRegionFromCache` to retrieve the region from the cache. If the region is not found in the cache, we then call `findRegion` to look up the region from the meta table. It seems that we are not establishing a connection when calling `getRegionFromCache`, but we are establishing one in `findRegion`. This leads me to think that we should establish a connection when loading a region into the cache. But I could be missing something

aaronbee: @dethi's suggestion is a good one, to make the connection establishment lazy, but I don't think it would work well today.
What you could do is remove the `MarkUnavailable()` and `establishRegion` calls. Then when an RPC is made `getRegionAndClientForRPC` will see that `reg.Client() == nil` for that region and then call `reestablishRegion`. The problem is that the way this works today is it calls `establishRegion` with `""` as the `addr`, which results in re-looking up the region from the meta region. So, then we are back to a flurry of requests to the meta region during start up (though, only one per region).

I think we should come up with a better fix for this in a later change.

rpc.go
aaronbee: The overlaps get `MarkDead()` called on them inside `*keyRegionCache.put`.

integration_test.go
```diff
@@ -2374,3 +2374,40 @@ func TestDebugState(t *testing.T) {
...
+	cacheLength := len(jsonUnMarshal["KeyRegionCache"].(map[string]interface{}))
+	if cacheLength != 0 {
+		t.Fatal("expected empty region cache when creating a new client")
+	}
+
+	// load regions to cache
+	c.CacheRegions([]byte(table))
+
+	jsonVal, err = gohbase.DebugState(c)
+	err = json.Unmarshal(jsonVal, &jsonUnMarshal)
```
aaronbee, line 2404: It would be best to use a new variable here rather than reuse `jsonUnMarshal`, since the effect is merging the results from the first `json.Unmarshal` and the second.

dethi:
Looks like there was a data race detected in the test, related to marshalJSON

rpc.go
dethi: sgtm, but need to get the CI to pass first

integration_test.go
```diff
@@ -2374,3 +2374,44 @@ func TestDebugState(t *testing.T) {
...
+	}
+	cacheLength := len(jsonUnMarshalStart["KeyRegionCache"].(map[string]interface{}))
+	if cacheLength != 0 {
+		t.Fatal("expected empty region cache when creating a new client")
+	}
+
+	c.CacheRegions([]byte(table))
+	// This ensures CacheRegions function completes execution before
+	// invoking DebugState, addressing potential data race issues
+	time.Sleep(time.Minute)
```
aaronbee, line 2403: This is not a great fix for the issue. It looks like the DebugState() code is really racy.

PR #254, opened 2024-06-04
hrpc: enable ScanMetrics

ciacono:
Enable tracking scan metrics in the ScanResponse. Clients can access the metrics via calls to scanner.Next()
Support for  HBase versions < 2.6.0 where ScanMetrics provides ROWS_SCANNED and ROWS_FILTERED metrics.


integration_test.go
```diff
@@ -1020,6 +1019,102 @@ func TestScanTimeRangeVersions(t *testing.T) {
...
+			expectedRowsScanned:  0,
+			expectedRowsFiltered: 2,
+		},
+	}
+
+	for _, tc := range tcases {
+		t.Run(tc.description, func(t *testing.T) {
+			var scan *hrpc.Scan
+			if tc.filters == nil {
+				scan, err = hrpc.NewScanStr(ctx, table, hrpc.TrackScanMetrics())
```
grigorecv, line 1083: check error?

scanner.go
```diff
@@ -32,16 +32,32 @@ type scanner struct {
...
+	var sm *hrpc.ScanMetrics
+	if s.rpc.TrackScanMetrics() {
+		sm = &hrpc.ScanMetrics{}
+	}
 	for {
 		resp, region, err := s.request()
+		if s.rpc.TrackScanMetrics() {
+			metrics := resp.ScanMetrics.GetMetrics()
+			for _, m := range metrics {
+				if m.GetName() == "ROWS_SCANNED" {
```
grigorecv, line 50: should we not have these as constants? (ROWS_FILTERED, ROWS_FILTERED)

integration_test.go
```diff
@@ -1020,6 +1019,102 @@ func TestScanTimeRangeVersions(t *testing.T) {
 	if len(rsp[1].Cells) != 1 {
 		t.Fatalf("Expected versions: %d, Got versions: %d", 2, len(rsp[0].Cells))
 	}
+
+	if rsp[0].ScanMetrics != nil && rsp[1].ScanMetrics != nil {
+		t.Fatal("Result had ScanMetrics when they were not enabled")
```
grigorecv, line 1024: could there be a test case for this as well?

scanner.go
```diff
@@ -51,29 +67,30 @@ func (s *scanner) fetch() ([]*pb.Result, error) {
...
-		rs, err := s.fetch()
+		rs, sm, err := s.fetch()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		// fetch cannot return zero results
 		s.results = rs
+		s.metrics = sm
```
grigorecv, line 91: Not sure s.metrics should be altered by peek(), maybe just return sm and not override s.metrics?

And a general question, should s.metrics ever be allowed to be changed inside a call of type func(s *scanner)?

grigorecv (approved):

pb/Client.proto
```diff
@@ -323,17 +323,21 @@ message ScanResponse {
...
   // server would have continued to scan had the time limit not been reached.
   optional bool heartbeat_message = 9;
-  
+
   // This field is filled in if the client has requested that scan metrics be tracked.
-  // The metrics tracked here are sent back to the client to be tracked together with 
+  // The metrics tracked here are sent back to the client to be tracked together with
   // the existing client side metrics.
-  //optional ScanMetrics scan_metrics = 10; // not in gohbase
+  optional ScanMetrics scan_metrics = 10; // not in gohbase
```
dethi, line 336: Can you remove the comment `// not in gohbase` and regenerate the pb.go?

scanner.go
```diff
@@ -18,7 +18,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-const noScannerID = math.MaxUint64
+const (
+	noScannerID  = math.MaxUint64
```
dethi, line 22: nit: empty space after since these constants have nothing to do with each other

scanner.go
```diff
@@ -35,13 +39,28 @@ type scanner struct {
...
+	var sm *hrpc.ScanMetrics
+	if s.rpc.TrackScanMetrics() {
+		sm = &hrpc.ScanMetrics{}
+	}
 	for {
 		resp, region, err := s.request()
+		if s.rpc.TrackScanMetrics() {
+			metrics := resp.ScanMetrics.GetMetrics()
+			for _, m := range metrics {
+				if m.GetName() == rowsScanned {
```
dethi, line 53: nit: use a switch statement instead of `if`

scanner.go
```diff
@@ -51,29 +70,35 @@ func (s *scanner) fetch() ([]*pb.Result, error) {
...
-		rs, err := s.fetch()
+		rs, sm, err = s.fetch()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 
 		// fetch cannot return zero results
 		s.results = rs
+		//s.metrics = sm
```
dethi, line 99: Commented code should be removed

dethi (commented):
Left a few comments but I also wonder if adding the scan metrics to the hrpc.Results is the right choice here. It seems that we will be allocating a lot of small struct on the heap to hold these result and most likely the user of this feature wants to know the overall scan metrics at the end so they will have to sum the various result in their own code.

Would it make more sense to hold the scan metrics in the scanner struct itself, accumulating the result as we fetch and add a method to the scanner to returns the end metrics. And also doing that while keepting the scan metrics as map[string]int64 such that there is no dependencies with the HBase version at all (if a new version of HBase introduce a new scan metrics name, or change one, we don't need to update GoHBase accordintly).

What are your thoughts on this?

scanner.go
```diff
@@ -30,15 +36,40 @@ type scanner struct {
...
 }
 
 func (s *scanner) fetch() ([]*pb.Result, error) {
 	// keep looping until we have error, some non-empty result or until close
 	for {
 		resp, region, err := s.request()
+		if s.rpc.TrackScanMetrics() && resp.ScanMetrics != nil {
+			metrics := resp.ScanMetrics.GetMetrics()
+			s.Lock()
+			for _, m := range metrics {
```
dethi, line 53: The for-loop can be replaced with just:
```go
for _, m := range metrics {
    s.scanMetrics[m.GetName()] += m.GetValue()
}
```
This way:
- no dependency between HBase and gohbase version
- `+=` on a map works since doing a map access on a key that doesn't already exist return the default value which for uint64 is 0

scanner.go
```diff
@@ -259,6 +286,19 @@ func (s *scanner) Close() error {
...
+// The scan metrics are non-nil only if the Scan has TrackScanMetrics() enabled
+// The scan metrics for a Scan are only accurate and complete if the scanner has
+// been closed with an io.EOF.
+func (s *scanner) GetScanMetrics() map[string]int64 {
+	if s.scanMetrics == nil {
+		return nil
+	}
+	s.RLock()
+	defer s.RUnlock()
+	return maps.Clone(s.scanMetrics)
```
aaronbee, line 299: This seems to allow getting the metrics before the scan is complete. Is that useful? Or could we get away with removing the locking around the metrics?

dethi (approved):
LGTM 🚢

PR #261, opened 2024-06-28
Fix logging to properly convert byte array to string

xdanica:
Properly convert []byte to string and add Client field in addition to client pointer

### DIFF
--- a/region/info.go
+++ b/region/info.go
@@ -12,6 +12,7 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"sync"
 
 	"github.com/tsuna/gohbase/hrpc"
@@ -358,10 +359,13 @@ func findCommaFromEnd(b []byte, offset int) int {
 
 func (i *info) MarshalJSON() ([]byte, error) {
 
-	var ctxError string
+	var ctxError, client string
 
 	if i.ctx != nil {
 		ctxError = fmt.Sprint(i.ctx.Err())
+	}
+	if i.Client() != nil {
+		client = i.Client().String()
 	}
 
 	state := struct {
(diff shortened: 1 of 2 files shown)


region/info.go
```diff
@@ -373,18 +376,20 @@ func (i *info) MarshalJSON() ([]byte, error) {
...
 		Client          string
 		Available       bool
 	}{
 		Id:              i.id,
-		Namespace:       string(i.namespace),
-		Table:           string(i.table),
-		Name:            string(i.name),
-		StartKey:        string(i.startKey),
-		StopKey:         string(i.stopKey),
+		Namespace:       fmt.Sprintf("%q", i.namespace),
```
aaronbee, line 384: I'd rather we use `strconv.QuoteToASCII`, because it's a little faster than going through `Sprintf`. And I think it deals better with unicode for our uses, ie. it should print out each byte instead of the unicode character.

One step better would to be write our own escape function that did exactly what we wanted, escaping non-ASCII characters rather than quoting the string. Not fully tested, but I think it works:
```
func escapeToASCII(in []byte) string {
	size := len(in)
	for _, c := range in {
		if c <= 0x20 || c >= 0x7f {
			size += 3
		} else if c == '\\' {
			size++
		}
	}
	const hexChars = "0123456789abcdef"
	out := make([]byte, 0, size)
	for _, c := range in {
		if c <= 0x20 || c >= 0x7f {
			out = append(out, '\\', 'x', hexChars[c>>4], hexChars[c&0xf])
		} else if c == '\\' {
			out = append(out, '\\', '\\')
		} else {
			out = append(out, c)
		}
	}
	return string(out)
}
```

xdanica: I used `strconv.QuoteToASCII` here: 
https://github.com/tsuna/gohbase/compare/1448caeca1d21bfcdc19407fee7aa1e35b61a5de..b49d419e5caad70bac8dfade1468f6e788c57c15

We could add a library as a follow up to add escapeToASCII, as well as the testing

aaronbee (approved):

PR #263, opened 2024-07-05
hrpc: add an option to skip retry

dethi:
This let the caller tell GoHBase to no retry calls that failed. Otherwise, GoHBase will retry forever until cancellation of the context. But sometime, it is useful to to be able to say "try only once", without just relying on having a small context timeout.

One of the place where this is useful is when we try to close a scanner that we prematurely stopped, as this is just the client trying to be a good citizen. But most likely, if the request failed with a retryable error (region closed or moved, call queue too big, etc), by the time we manage to send the request the scanner lease would have been already expired on HBase side.


dethi:
TODO:
- [ ] New unit test
- [ ] Documentation

But open to feedback about the API itself, if someone think we should do this differently or not at all?

PR #267, opened 2024-08-14
hrpc, region: Allow setting priority on Scan and Get requests

aaronbee:
Provide an option for setting the priority on a Scan or Get request. The priority is set on the RequestHeader, so must be set by the region code when serializing the request.

### DIFF
--- a/hrpc/query.go
+++ b/hrpc/query.go
@@ -23,6 +23,7 @@ type baseQuery struct {
 	maxVersions   uint32
 	storeLimit    uint32
 	storeOffset   uint32
+	priority      uint32
 	cacheBlocks   bool
 	consistency   ConsistencyType
 }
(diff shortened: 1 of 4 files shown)


dethi (approved):

hrpc/query.go
```diff
@@ -97,6 +98,15 @@ func (bq *baseQuery) setCacheBlocks(cacheBlocks bool) {
 func (bq *baseQuery) setConsistency(consistency ConsistencyType) {
 	bq.consistency = consistency
 }
+func (bq *baseQuery) setPriority(priority uint32) {
+	bq.priority = priority
+}
+func (bq *baseQuery) Priority() *uint32 {
```
tsuna, line 104: Returning by point here is weird. I understand that when we marshal to proto it wants pointers-to-int and maybe doing it this way avoids an allocation (because otherwise in `marshalProto()` you'd have to create an `int` and take its address thereby causing a heap allocation of a new integer), but from an API perspective it's odd. Did you do it this way to avoid an allocation?

aaronbee: Yeah, to avoid an allocation.

aaronbee: And now this has become slightly awkward in #268, which fixes something I missed here.

dethi: Oh I didn't notice that. I think returning just uint32 make more sense. Returning a pointer or a uint32 is the same in the end

aaronbee: It's same except for the added allocation. The extra allocation is not a big deal, so I'll just clean this up in #268.

aaronbee: Updated #268

PR #273, opened 2024-10-22
rpc: measure the amount of time spend sleeping in retry backoff

dethi:
This information is useful to understand if the HBase clusters is overloaded (CallQueueTooBigException) and track the impact of it on gohbase latency.


aaronbee (approved):
Thanks!

PR #274, opened 2024-10-22
rpc: ignore offline/unassigned regions during cache warmup

dethi:
When warming up the cache, ignore regions that are offline/unassigned. They could exist for many reasons:
- parent region after a split, until daughter regions are compacted
- regions that are merged together, until compaction happens
- region moving due to HBase LoadBalancer, RegionServer shutting down or manually
- etc

It's not critical if we are missing a few regions in the cache during warmup. The most important thing is for the warmup to be fast and to fill in with as many regions as possible. Other regions will be fetched later as needed.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -757,7 +757,7 @@ func createAllRegionSearchKey(table []byte) []byte {
 	return metaKey
 }
 
-// metaLookupForTable checks meta table for all the region in which the given table is.
+// metaLookupForTable checks meta table for all the open table regions.
 func (c *client) metaLookupForTable(ctx context.Context,
 	table []byte) ([]regionInfoAndAddr, error) {
 	metaKey := createAllRegionSearchKey(table)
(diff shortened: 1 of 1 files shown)


rpc.go
```diff
@@ -780,7 +780,16 @@ func (c *client) metaLookupForTable(ctx context.Context,
...
+			// shouldn't happen unless HBase meta table is corrupted/changed format.
+			if _, ok := err.(region.OfflineRegionError); !ok {
+				c.logger.Debug("failed to parse region", "err", err)
+			}
+			continue
+		}
+		if addr == "" {
+			// Ignore region that aren't open/assigned to a RegionServer.
+			continue
 		}
```
gogochickenleg, line 793: I think we are already doing this check in `ParseRegionInfo`
https://github.com/tsuna/gohbase/blob/ea1d2bc61069aff06b0eb4aa7f449e99c0f5c225/region/info.go#L150

dethi: How did I miss that lol

PR #276, opened 2024-11-19
AdminClient: set snapshot table and apply snapshot to new table

ciacono:
Add method to change the table of the hrpc.Snapshot. By setting the snapshot to a new table name that doesn't already exist, the data in the snapshot can be cloned into this new table instead of being restored to the table the snapshot was created from.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -2279,9 +2279,9 @@ func TestSnapshot(t *testing.T) {
 	}
 }
 
-// Test snapshot restoration
+// TestRestoreSnapshot tests using a snapshot to restore a table.
 func TestRestoreSnapshot(t *testing.T) {
-	// Prochedure for this test is roughly:
+	// Procedure for this test is roughly:
 	// - Create some data in a table.
 	// - Create a snapshot.
 	// - Remove all data.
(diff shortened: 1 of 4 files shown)


integration_test.go
```diff
@@ -2515,3 +2515,122 @@ func TestCacheRegions(t *testing.T) {
...
+
+	// Name of new table to restore snapshot data to.
+	// The new table is not created at this time.
+	// HBase will do this when trying to restore a snapshot to a table that does not already exist.
+	// If the snapshot table doesn't exist, as in this case, HBase will clone the snapshot to a new
+	// table.
+	tableNew := fmt.Sprintf("gohbase_test_%d_%s", time.Now().UnixNano(), t.Name())
+
+	// Update snapshot to new table:
+	sn.SetTable(tableNew)
```
dethi, line 2580: Do we really a new method? It feels like to me doing `sn, err := hrpc.NewSnapshot(context.Background(), snapshotName, tableNew); [...] ac.RestoreSnapshot(sn)` is enough (i.e just create a new snapshot object with the same name but a new table name)

ciacono: Done!

dethi (approved):

PR #277, opened 2024-12-03
add renewal for scanner to prevent lease timeout

tanay-vakharia:
This change adds an option to Scan (Renew) that allows for renewal of scanners during periods of starvation to prevent lease timeout. Currently, we are renewing every lease timeout / 2 seconds (10s currently).

### DIFF
--- a/hrpc/scan.go
+++ b/hrpc/scan.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"time"
 
 	"github.com/tsuna/gohbase/pb"
 	"google.golang.org/protobuf/proto"
@@ -77,6 +78,9 @@ type Scan struct {
 
 	closeScanner        bool
 	allowPartialResults bool
+
+	renewInterval time.Duration
+	renewalScan   bool
 }
 
 // baseScan returns a Scan struct with default values set.
@@ -93,6 +97,8 @@ func baseScan(ctx context.Context, table []byte,
 		maxResultSize: DefaultMaxResultSize,
 		numberOfRows:  DefaultNumberOfRows,
 		reversed:      false,
+		renewInterval: 0 * time.Second,
+		renewalScan:   false,
 	}
 	err := applyOptions(s, options...)
 	if err != nil {
@@ -104,10 +110,12 @@ func baseScan(ctx context.Context, table []byte,
 func (s *Scan) String() string {
 	return fmt.Sprintf("Scan{Table=%q StartRow=%q StopRow=%q TimeRange=(%d, %d) "+
 		"MaxVersions=%d NumberOfRows=%d MaxResultSize=%d Familes=%v Filter=%v "+
-		"StoreLimit=%d StoreOffset=%d ScannerID=%d Close=%v}",
+		"StoreLimit=%d StoreOffset=%d ScannerID=%d Close=%v RenewInterval=%v"+
+		"RenewalScan=%v}",
 		s.table, s.startRow, s.stopRow, s.fromTimestamp, s.toTimestamp,
 		s.maxVersions, s.numberOfRows, s.maxResultSize, s.families, s.filter,
-		s.storeLimit, s.storeOffset, s.scannerID, s.closeScanner)
+		s.storeLimit, s.storeOffset, s.scannerID, s.closeScanner, s.renewInterval,
+		s.renewalScan)
 }
 
 // NewScan creates a scanner for the given table.
@@ -189,6 +197,18 @@ func (s *Scan) TrackScanMetrics() bool {
 	return s.trackScanMetrics
 }
 
+// RenewInterval returns the interval at which the scanner will be renewed
+// which is usually lease timeout / 2 secs
+func (s *Scan) RenewInterval() time.Duration {
+	return s.renewInterval
+}
+
+// RenewalScan returns whether this scan is to be used only a renewal request
+// to hbase
+func (s *Scan) RenewalScan() bool {
+	return s.renewalScan
+}
+
 // ToProto converts this Scan into a protobuf message
 func (s *Scan) ToProto() proto.Message {
 	scan := &pb.ScanRequest{
@@ -201,6 +221,11 @@ func (s *Scan) ToProto() proto.Message {
 		// since we don't really time out our scans (unless context was cancelled)
 		ClientHandlesHeartbeats: proto.Bool(true),
 		TrackScanMetrics:        &s.trackScanMetrics,
+		Renew:                   proto.Bool(false),
+	}
+	// Tells hbase whether this request is for scanner renewal
+	if s.renewalScan {
+		scan.Renew = &s.renewalScan
 	}
 	if s.scannerID != math.MaxUint64 {
 		scan.ScannerId = &s.scannerID
@@ -384,6 +409,33 @@ func Attribute(key string, val []byte) func(Call) error {
 			return errors.New("'Attributes' option can only be used with Scan queries")
 		}
 		scan.attribute = append(scan.attribute, &pb.NameBytesPair{Name: &key, Value: val})
+		return nil
+	}
+}
+
+// RenewInterval is a an option for scan requests.
+// Enables renewal of scanners at an interval to prevent timeout of scanners due to
+// waiting/starvation
+func RenewInterval(interval time.Duration) func(Call) error {
+	return func(g Call) error {
+		scan, ok := g.(*Scan)
+		if !ok {
+			return errors.New("'RenewInterval' option can only be used with Scan queries")
+		}
+		scan.renewInterval = interval
+		return nil
+	}
+}
+
+// RenewalScan is an option for scan requests.
+// Indicates that this Scan request will be used for the renewal of a scanner only
+func RenewalScan() func(Call) error {
+	return func(g Call) error {
+		scan, ok := g.(*Scan)
+		if !ok {
+			return errors.New("'RenewScan' option can only be used with Scan queries")
+		}
+		scan.renewalScan = true
 		return nil
 	}
 }
--- a/integration_test.go
+++ b/integration_test.go
@@ -44,6 +44,8 @@ func init() {
 	table = fmt.Sprintf("gohbase_test_%d", time.Now().UnixNano())
 }
 
+const scannerLease = 5 * time.Second
+
 // CreateTable creates the given table with the given families
 func CreateTable(client gohbase.AdminClient, table string, cFamilies []string) error {
 	// If the table exists, delete it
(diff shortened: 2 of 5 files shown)


integration_test.go
```diff
@@ -1094,6 +1094,7 @@ func TestScanWithScanMetrics(t *testing.T) {
 			for {
 				var r *hrpc.Result
 				r, err = scanner.Next()
+				fmt.Println("NEXT CALL DONE")
```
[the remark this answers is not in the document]
tanay-vakharia: Removed. Thanks!

hrpc/scan.go
```diff
@@ -387,3 +399,16 @@ func Attribute(key string, val []byte) func(Call) error {
 		return nil
 	}
 }
+
+// Renew is a an option for scan requests.
+// Enables renewal of scanners at an interval to prevent timeout of scanners due to
```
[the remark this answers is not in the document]
tanay-vakharia: Finished the comment, thanks for pointing that out!

integration_test.go
```diff
@@ -1058,17 +1058,17 @@ func TestScanWithScanMetrics(t *testing.T) {
...
-			expectedRowsScanned:  2,
-			expectedRowsFiltered: 0,
-		},
-		{
-			description:          "1 row scanned 1 row filtered",
-			filters:              hrpc.Filters(filter.NewPrefixFilter([]byte(r1))),
-			expectedRowsScanned:  1,
-			expectedRowsFiltered: 1,
-		},
+		// {
```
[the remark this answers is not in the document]
tanay-vakharia: Sorry, this was leftover from when I was testing my code, I uncommented it and it should be the way it was before now.

integration_test.go
```diff
@@ -2634,3 +2634,118 @@ func TestNewTableFromSnapshot(t *testing.T) {
 		t.Fatalf("expected no cells after RestoreSnapshot in table %s key %s", table, key)
 	}
 }
+
+// TestScannerTimeout tests whether there is a lease timeout between Next calls
+// if the wait between them is too long. This is fixed by the renewal process
```
gogochickenleg, line 2639: nit: I think the purpose of this test is to make sure when the Renew() flag is off, we are indeed getting the UnknownScannerException error. So maybe change the comment to something like 
```
// TestScannerTimeout makes sure that without the Renew flag on, we get a lease timeout between Next calls
// if the wait between them is too long. 
```

scanner.go
```diff
@@ -137,6 +140,8 @@ func newScanner(c RPCClient, rpc *hrpc.Scan) *scanner {
 		startRow:           rpc.StartRow(),
 		curRegionScannerID: noScannerID,
 		scanMetrics:        sm,
+		// value is lease timeout / 2
+		renewInterval: 10 * time.Second,
```
gogochickenleg, line 144: I dont think we should hardcode this since gohbase is open source and different users might have different lease timeout set. The problem is that the lease timeout is an hbase flag that you configure when starting up hbase so we don't have access to that number here AFAIK.

One way that I can think of is adding a flag `renewInterval` to the hrpc.Scan struct to set the time interval there and pass it down to here. Or we can change the hrpc.Scan.renew flag to be a pointer to `time.Duration` and only use it if it's not nil. @aaronbee lmk if you have other ideas or if I am wrong

scanner.go
```diff
@@ -202,13 +216,26 @@ func (s *scanner) request() (*pb.ScanResponse, hrpc.RegionInfo, error) {
...
+		options := s.rpc.Options()
+		filteredOptions := make([]func(hrpc.Call) error, 0, len(options))
+		for _, opt := range options {
+			dummyScan := &hrpc.Scan{}
+			opt(dummyScan)
+			if !dummyScan.Renew() {
+				filteredOptions = append(filteredOptions, opt)
+			}
+		}
+
```
aaronbee, line 231: Hopefully not needed after implementing suggestion above.

hrpc/scan.go
```diff
@@ -33,6 +33,8 @@ const (
 	DefaultMaxResultsPerColumnFamily = math.MaxInt32
 	// DefaultCacheBlocks is the default setting to enable the block cache for get/scan queries
 	DefaultCacheBlocks = true
+	// DefaultRenew is the default setting to enable renewing for scanners
```
ciacono, line 36: Does this need to be defined as a const if the default is false?
vs. just at line 100 setting `renew: false,`

tanay-vakharia: I added this because I wasn't sure if the defaults ever needed to be changed in the future (for whatever reason) and I was following the same pattern as the other options. I can remove this and simply set it in the constructor if others agree.

integration_test.go
tanay-vakharia: Changed the comment, thanks!

scanner.go
tanay-vakharia: Implemented this with a slight change. Since converting to a protobuf is used internally within `SendRPC`, I split apart the `renew` boolean into 2, `renewInterval` which tracks the interval at which the scanner is renewed, and `renewalScan` which indicates whether this Scan request is a renewal request. These are both Scan options which can be set. Then in `ToProto` for `Scan` we check if `renewalScan` is true and if it is, then we set the `Renew` field in the protobuf messages accordingly. Let me know if this looks fine!

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
...
+		s.renewTimer.Stop()
+	}
+}
+
+// renews the scanner and restarts timer after renewal
+func (s *scanner) renewScanner() {
+	err := s.renew()
+	if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
tanay-vakharia: Logged it here!

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
 	}
 	s.curRegionScannerID = noScannerID
 }
+
+// renews a scanner by resending scan request with renew = true
+func (s *scanner) renew() error {
+	select {
+	case <-s.rpc.Context().Done():
+		return errors.New("context was canceled")
+	default:
```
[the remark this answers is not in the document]
tanay-vakharia: Changed it to this!

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
...
+		}
+		// save new scanner id
+		s.curRegionScannerID = resp.GetScannerId()
+		return nil
+	}
+}
+
+// starts the renew timer, after it fires, renewScanner is called
+func (s *scanner) startRenewTimer() {
+	if s.rpc.Renew() {
```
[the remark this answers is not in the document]
tanay-vakharia: Added a check for whether the timer is nil at the beginning. It can be an invariant that before calling `startRenewTimer`, the `renewTimer` field must be nil. To follow that invariant, I also stopped the renew timer as well at the end of `renewScanner`.

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
...
+		s.renewTimer = time.AfterFunc(s.renewInterval, s.renewScanner)
+	}
+}
+
+// stops the renewal timer
+func (s *scanner) stopRenewTimer() {
+	if s.rpc.Renew() && s.renewTimer != nil {
+		s.renewTimer.Stop()
```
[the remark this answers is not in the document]
tanay-vakharia: Set!

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
...
+// starts the renew timer, after it fires, renewScanner is called
+func (s *scanner) startRenewTimer() {
+	if s.rpc.Renew() {
+		s.renewTimer = time.AfterFunc(s.renewInterval, s.renewScanner)
+	}
+}
+
+// stops the renewal timer
+func (s *scanner) stopRenewTimer() {
+	if s.rpc.Renew() && s.renewTimer != nil {
```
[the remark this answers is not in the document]
tanay-vakharia: Fixed this, thanks!

hrpc/scan.go
gogochickenleg: Both approach seems fine to me.

>I added this because I wasn't sure if the defaults ever needed to be changed in the future

Good point, I think it'd make sense to define DefaultRenew explicitly if this is something that we might change/tune in the future. But in this case, I think we are unlikely to change it, so hard coding `false` on line 100 seems like a slightly better approach.

scanner.go
tanay-vakharia: Removed now!

scanner.go
```diff
@@ -149,6 +154,15 @@ func toLocalResult(r *pb.Result) *hrpc.Result {
 }
 
 func (s *scanner) Next() (*hrpc.Result, error) {
+	s.stopRenewTimer()
+	res, err := s.nextInternal()
+	if err == nil && s.rpc.Renew() {
+		s.startRenewTimer()
+	}
```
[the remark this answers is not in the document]
tanay-vakharia: Agree, now the checks are only before calling the start and stop functions.

scanner.go
```diff
@@ -359,3 +387,58 @@ func (s *scanner) closeRegionScanner() {
...
+			nil,
+			hrpc.ScannerID(s.curRegionScannerID),
+			hrpc.NumberOfRows(s.rpc.NumberOfRows()),
+			hrpc.Priority(s.rpc.Priority()),
+			hrpc.Renew(),
+		)
+		if err != nil {
+			return err
+		}
+		res, sendErr := s.SendRPC(rpc)
```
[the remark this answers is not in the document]
tanay-vakharia: Fixed!

hrpc/scan.go
tanay-vakharia: Ok sounds good, I've hard coded 0 seconds for `RenewInterval` and false for `RenewalScan` in `baseScan`.

scanner.go
```diff
@@ -359,3 +378,63 @@ func (s *scanner) closeRegionScanner() {
...
+	}
+}
+
+// starts the renew timer, after it fires, renewScanner is called
+func (s *scanner) startRenewTimer() {
+	if s.renewTimer != nil {
+		panic("Renew timer was not stopped before starting again")
+	}
+	if s.rpc.RenewInterval() > 0 {
+		s.renewTimer = time.AfterFunc(s.rpc.RenewInterval(), s.renewScanner)
```
aaronbee, line 419: A different way to do this would be to start a goroutine that sleeps on a ticker and context.

```
renewCtx, cancel := context.WithCancel(s.Context())
s.renewCancel = cancel
go s.renewLoop(renewCtx)
```

And `renewLoop` would look like:
```
func (s *scanner) renewLoop() {
    t := time.Ticker(s.rpc.RenewInterval()
    defer t.Stop()
    for {
        select {
        case <-t.C:
             if err := s.renew(); err != nil {
                    return
              }
        case <-ctx.Done():
            return
        }
    }
}
```

The advantage here is that the renewLoop routine will clean itself up immediately after the context is canceled and doesn't have to wait for the timer to go off to notice the context is done. And I think it's slightly simpler to have a loop rather than `renewScanner()` needing to start a new renew timer.

scanner.go
```diff
@@ -359,3 +378,63 @@ func (s *scanner) closeRegionScanner() {
...
+		s.renewTimer.Stop()
+		s.renewTimer = nil
+	}
+}
+
+// renews the scanner and restarts timer after renewal
+func (s *scanner) renewScanner() {
+	err := s.renew()
+	if err != nil {
+		s.logger.Debug("Error in renewing: ", slog.String("err: ", err.Error()))
```
aaronbee, line 435: This log at the error level. And because this is a structured logger you don't need to include ": " in your strings. So more like:

`s.logger.Error("error renewing scanner", "err", err)`

scanner.go
```diff
@@ -359,3 +378,63 @@ func (s *scanner) closeRegionScanner() {
...
+func (s *scanner) renew() error {
+	if err := s.rpc.Context().Err(); err != nil {
+		return err
+	} else {
+		rpc, err := hrpc.NewScanRange(s.rpc.Context(),
+			s.rpc.Table(),
+			s.startRow,
+			nil,
+			hrpc.ScannerID(s.curRegionScannerID),
+			hrpc.NumberOfRows(s.rpc.NumberOfRows()),
```
aaronbee, line 392: These shouldn't be needed on the RenewScan.

timoha:
Just my 2c:
What's preventing scanning go routine to keep scanning and storing in a buffer end purposefully renewing the lease? I think the lease should only be renewed purposefully if you are actively scanning data.

Also, please make sure to not have this behavior by default as it might be unexpected for some users, as now their "periods of starvation" in client layer will cause leaked snanners within regionservers.

scanner.go
```diff
@@ -359,3 +378,63 @@ func (s *scanner) closeRegionScanner() {
...
 }
+
+// renews a scanner by resending scan request with renew = true
+func (s *scanner) renew() error {
+	if err := s.rpc.Context().Err(); err != nil {
+		return err
+	} else {
+		rpc, err := hrpc.NewScanRange(s.rpc.Context(),
+			s.rpc.Table(),
+			s.startRow,
```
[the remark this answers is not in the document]
aaronbee: Ughh... I just realized this doesn't work because gohbase doesn't allow you to send the same rpc twice. 
So, the way to do this would be to pass in the startRow.

aaronbee:
> Just my 2c: What's preventing scanning go routine to keep scanning and storing in a buffer end purposefully renewing the lease? I think the lease should only be renewed purposefully if you are actively scanning data.
> 
> Also, please make sure to not have this behavior by default as it might be unexpected for some users, as now their "periods of starvation" in client layer will cause leaked snanners within regionservers.

Hey Andrey!
Yeah, I agree this shouldn't be enabled by default inside gohbase.
We have some needs for keeping a long lived scanner running. Buffering extra data is not great because of the extra memory use.

timoha:
Hey Aaron!
Got it, it makes sense as an option. Just another alternative I'd suggest in those cases is just increasing lease timeout (server or client config). I would expect that eventually you would have to timeout anyway with your scan

scanner.go
tanay-vakharia: Tested this out, we need the scannerId to identify the scanner we need to renew, but we don't need the number of rows so that will be removed.

scanner.go
aaronbee: Another advantage here is that it provides a place for us to add a metric to track the number of live renewals. It can be done in a follow on change, but we will want to add a metric that increments when this function starts and decrements when it exits. That way we can track if we are leaking scanners.

scanner.go
aaronbee: Oh, right. Of course the ScannerID is needed 🤦

scanner.go
tanay-vakharia: Thanks, this definitely cleans up the code and makes it easier to add metrics in the future. I've implemented this, feel free to take a look!

scanner.go
tanay-vakharia: Fixed!

scanner.go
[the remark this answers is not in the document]
tanay-vakharia: Implemented this, let me know if you have any feedback!

integration_test.go
```diff
@@ -2634,3 +2634,180 @@ func TestNewTableFromSnapshot(t *testing.T) {
...
+	expectedValue := []byte(strconv.Itoa(0))
+	if !bytes.Equal(rsp.Cells[0].Value, expectedValue) {
+		t.Errorf("Unexpected value. Got %v, want %v", rsp.Cells[0].Value, expectedValue)
+	}
+
+	// Cancel the context
+	cancel()
+
+	// Sleep to allow time for cancellation to propagate
+	time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
tanay-vakharia: Fixed.

aaronbee (approved):

PR #281, opened 2024-12-30
Don't restart scanner renewer on every Next call

aaronbee:
Most calls to Next are just reading some buffered results. The scanner renewer only needs to be stopped and started again when we ask for more results from HBase. Also, we shouldn't start a renewer if we know we have already fetched the final results.

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -75,24 +75,35 @@ func (s *scanner) fetch() ([]*pb.Result, error) {
 }
 
 func (s *scanner) peek() (*pb.Result, error) {
-	if len(s.results) == 0 {
-		var (
-			err error
-			rs  []*pb.Result
-		)
-		if s.closed {
-			// done scanning
-			return nil, io.EOF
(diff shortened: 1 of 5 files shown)


scanner.go
```diff
@@ -75,24 +75,35 @@ func (s *scanner) fetch() ([]*pb.Result, error) {
...
+	}
 
-		// fetch cannot return zero results
-		s.results = rs
+	if s.renewCancel != nil {
+		// About to send new Scan request to HBase, cancel our
+		// renewer.
+		s.renewCancel()
+		s.renewCancel = nil
+	}
```
gogochickenleg, line 92: Should we swap the order of these two if clauses? to ensure we cancel the renewer before returning EOF

gogochickenleg (commented):
LGTM!

scanner.go
aaronbee: Done.

PR #289, opened 2025-03-10
root: Remove Redundant Type Casting

cluckett-arista:
There are several places with redundant type casting. Presumably, many if not all of these were probably due to constraints of the early Go compiler. Although there is a certain readability aspect, there is potential over- head in redundant casts. The Go compiler might add in overhead to cast (e.g., `string` to `[]byte`). These were found with GoLand's linter.

### DIFF
--- a/region/compressor_test.go
+++ b/region/compressor_test.go
@@ -181,7 +181,7 @@ func benchmarkDecompressCellblocks(b *testing.B, blockLen int, blocksCount int) 
 			b.FailNow()
 		}
 
-		cellblocks := c.compressCellblocks([][]byte{data}, uint32(blockLen))
+		cellblocks := c.compressCellblocks(net.Buffers{data}, uint32(blockLen))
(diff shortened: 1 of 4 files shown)


cluckett-arista:
Other options may include being more selective with the conversions in order to measure impact. There are also times for redundant conversions (readability).

region/compressor_test.go
```diff
@@ -181,7 +181,7 @@ func benchmarkDecompressCellblocks(b *testing.B, blockLen int, blocksCount int)
 			b.FailNow()
 		}
 
-		cellblocks := c.compressCellblocks(net.Buffers([][]byte{data}), uint32(blockLen))
+		cellblocks := c.compressCellblocks([][]byte{data}, uint32(blockLen))
```
cluckett-arista, line 184: This is a mistake. The alias should be used here.

dethi (approved):

PR #290, opened 2025-03-27
Root: Remove Old Build Tags

cluckett-arista:
Old-style build tags no longer needed. They can be removed. Sorry, they were bugging me.


ciacono (commented):
There's one more old tag in `region/new.go`. Do you want to remove that in this PR as well for completeness?

ciacono (approved):

PR #291, opened 2025-03-27
Root: Add Nix-Shell File

cluckett-arista:
Adds a nix-shell file for dev support while working in the repo.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,28 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.mkShell {
+
+  packages = with pkgs; [
+    hbase_2_4
+    jdk8
+    hadoop
+    go_1_24
+    procps
+  ];
+
+  shellHook = ''
+    export JAVA_HOME="${pkgs.jdk8}"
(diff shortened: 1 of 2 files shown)


shell.nix
```diff
@@ -0,0 +1,28 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.mkShell {
+
+  packages = with pkgs; [
+    hbase_2_4
+    jdk8
+    hadoop
+    go_1_24
+    procps
```
cluckett-arista, line 10: sorry, this was for debugging. I can either add that back, so this is needed again, or I can remove it.

shell.nix
```diff
@@ -0,0 +1,33 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+pkgs.mkShell {
+
+  packages = with pkgs; [
+    jdk8    
+    hbase_2_4
+    hadoop
+
+    go_1_24
```
ciacono, line 10: The current go version for gohbase in go.mod is 1.23.0, should be kept in sync?

PR #292, opened 2025-04-11
pb: Format .proto Files

cluckett-arista:
Formats (using GoLand) to avoid merge conflicts at a later time.


dethi:
These are copy/pasted from HBase code base. Please don't reformat them, it's pointless.

PR #293, opened 2025-04-17
scan: add ScanStatsHandler and ScanStats

ciacono:
Create a new scan option to provide a ScanStatsHandler function that will be called when a scan is completed. The ScanStatsHandler is called with ScanStats as an arg. ScanStats is updated, and the client's ScanStatsHandler is called to use the ScanStats as desired.


scanner.go
```diff
@@ -88,11 +92,19 @@ func (s *scanner) peek() (*pb.Result, error) {
 
 	if s.closed {
 		// done scanning
+		if s.rpc.ScanStatsHandler() != nil {
+			updateScanStats(s.rpc.ScanStats(), s.rpc, s, nil, time.Since(s.start))
+			s.rpc.ScanStatsHandler()(s.rpc.ScanStats())
+		}
```
aaronbee, line 98: I was imagining that we would emit a ScanStats for every single scan including the partial scans. ie. Emit one of these inside `request()` for every call to `SendRPC`. Going a step further I would like to be able to emit a ScanStats for every failure and retry inside `SendRPC`.

The reason to do this is that the sub-scans may end up going to different regions or regionservers, so to track that properly we need to emit a scanstats for each of those requests. Also, I wanted to capture the cases where we are hitting an overloaded region server, getting CallQueueTooBig exceptions and retrying.

hrpc/scan.go
```diff
@@ -209,6 +230,27 @@ func (s *Scan) RenewalScan() bool {
...
+	return s.scanStats
+}
+
+func (s *Scan) ColumnFamilies() [][]byte {
+	cfs := make([][]byte, len(s.families))
+	for f, _ := range s.families {
+		cfs = append(cfs, []byte(f))
+	}
+	return cfs
+}
```
aaronbee, line 247: I think we can leave out ColumnFamilies for now. My concern is that this is a little more expensive to add to the scan stats since it requires allocations and it could end up being a large slice.

rpc.go
```diff
@@ -126,12 +134,61 @@ func (c *client) SendRPC(rpc hrpc.Call) (msg proto.Message, err error) {
...
+
+		stats.Table = scan.Table()
+		stats.StartRow = scan.StartRow()
+		stats.EndRow = scan.StopRow()
+		if scan.Region() != nil {
+			stats.RegionID = scan.Region().ID()
+			if scan.Region().Client() != nil {
+				stats.RegionServer = scan.Region().Client().Addr()
+			}
+		}
```
aaronbee, line 180: A region's client can change in the background, so one call to `scan.Region().Client()` may return non-nil and then later return nil.

The results of `scan.Region()` (though, this one probably can't change in the background) and `scan.Region().Client()` should be saved into variables, ie.

```
if reg := scan.Region(); reg != nil {
    ...
    if c := reg.Client(); c != nil {
        ...
```

rpc.go
```diff
@@ -126,12 +134,61 @@ func (c *client) SendRPC(rpc hrpc.Call) (msg proto.Message, err error) {
...
+			if !ok {
+				c.logger.Error("got non ScanResponse for ScanRequest")
+			} else {
+				if scanres.ScanMetrics != nil {
+					if stats.ScanMetrics == nil {
+						stats.ScanMetrics = make(map[string]int64)
+					}
+					for _, m := range scanres.ScanMetrics.GetMetrics() {
+						stats.ScanMetrics[m.GetName()] += m.GetValue()
+					}
```
aaronbee, line 167: This is now being updated in two places, here and in `scanner.fetch`, which I think means the same metrics will be added twice. And as your comment above mentions this is being added up across multiple scans and for the ScanStats we should just have the metrics for one scan.
Perhaps you can just create a local `map[string]int64` for the metrics found for just this one scan and put that in the ScanStats struct. And leave the code in `scanner.fetch` to be responsible for summing the ScanMetrics across multiple scans.

ciacono: For the ScanStats, there is a local map[string]int64 for the ScanStats.ScanMetrics. This is different from the scanner.ScanMetrics that's updated in scanner.fetch. which sums the metrics. I've updated the ScanStats one to not add it like in scanner.fetch.

hrpc/scan_test.go
```diff
@@ -0,0 +1,83 @@
...
+		input.StartRow = []byte("start")
+		input.EndRow = []byte("end")
+		input.RegionID = 123
+		input.RegionServer = "a region server addr"
+		input.ScannerID = 456
+		input.ScanMetrics = map[string]int64{"METRIC_ONE": int64(777), "METRIC_TWO": int64(888)}
+		input.Start = time.Unix(0, 1)
+		input.End = time.Unix(0, 2)
+		input.Error = true
+		input.Retryable = false
```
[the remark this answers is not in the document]
ciacono: Removed from test and added some checks in rpc_test.go

hrpc/scan.go
```diff
@@ -439,3 +477,41 @@ func RenewalScan() func(Call) error {
 		return nil
 	}
 }
+
+// ScanStatsID is an option for Scan requests to provide a ScanStatsID for the scan
+func ScanStatsID(id int64) func(Call) error {
```
aaronbee, line 482: Can we include a comment here saying that this is just used internally by gohbase

aaronbee (approved):

PR #297, opened 2025-07-17
hrpc: guard against time range inputs > Long.MAX_VALUE

ciacono:


hrpc/mutate.go
```diff
@@ -100,6 +100,9 @@ func TimestampUint64(ts uint64) func(Call) error {
 		if !ok {
 			return errors.New("'TimestampUint64' option can only be used with mutation queries")
 		}
+		if ts > MaxTimestamp {
```
aaronbee, line 103: (GitHub is not allowing me to comment on lines that are not part of the diff, so I'm posting it here.)

In `Timestamp` you should also check the time. It's possible that the user passes in a time.Time that is before the unix epoch, which will end up being negative / a value greater than MaxTimestamp when converted to uint64.

hrpc/mutate.go
```diff
@@ -494,7 +497,8 @@ var durabilities = []*pb.MutationProto_Durability{
 
 func (m *Mutate) toProto(isCellblocks bool, cbs [][]byte) (*pb.MutateRequest, [][]byte, uint32) {
 	var ts *uint64
-	if m.timestamp != MaxTimestamp {
```
aaronbee, line 497: I think this was intentional behavior, though it looks odd. `m.timestamp` is set to MaxTimestamp (MaxUint64) by default during instantiation, which is meant to be interpreted as unset. So, that's why this code is not setting the pb value.
I think we may need to re-name `MaxTimestamp` to `UnsetTimestamp` and leave that as `math.MaxUint64` then make `MaxTimestamp` be `MaxInt64`.

PR #301, opened 2025-08-14
hrpc: Add CheckAndMutate

cluckett-arista:
Adds the wrapper for [CheckAndMutate HRPC calls]([https://hbase.apache.org/2.5/apidocs/org/apache/hadoop/hbase/client/CheckAndMutate.html]). Although general Mutation functionality somewhat exists through the CheckAndPut operation, [we don't allow any kind of mutations besides Puts](https://github.com/tsuna/gohbase/blob/master/hrpc/checkandput.go#L32). This PR introduces a new binding for `CheckAndMutate` that neatly exposes this new behavior.

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,6 +42,8 @@ type Client interface {
 	Increment(i *hrpc.Mutate) (int64, error)
 	CheckAndPut(p *hrpc.Mutate, family string, qualifier string,
 		expectedValue []byte) (bool, error)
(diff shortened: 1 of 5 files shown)


cluckett-arista:
Closed in favor of updating CheckAndPut (until HBase removes CheckAndPut in 3.0.0): https://github.com/tsuna/gohbase/pull/302

PR #302, opened 2025-08-14
hrpc: CheckAndPutWithCompareType

cluckett-arista:
HBase offers the ability to use more than only `CompareType_EQUAL`. This allows other `CompareType` values to be used.

### DIFF
--- a/hrpc/checkandput.go
+++ b/hrpc/checkandput.go
@@ -6,8 +6,6 @@
 package hrpc
 
 import (
-	"fmt"
-
 	"github.com/tsuna/gohbase/filter"
 	"github.com/tsuna/gohbase/pb"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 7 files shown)


hrpc/checkandput.go
```diff
@@ -21,14 +21,22 @@ type CheckAndPut struct {
...
-	comparator *pb.Comparator
+	compareType *pb.CompareType
+	comparator  *pb.Comparator
 }
 
-// NewCheckAndPut creates a new CheckAndPut request that will compare provided
-// expectedValue with the on in HBase located at put's row and provided family:qualifier,
-// and if they are equal, perform the provided put request on the row
+// NewCheckAndPut creates a CheckAndPut with only an EQUAL compare type.
 func NewCheckAndPut(put *Mutate, family string,
```
cluckett-arista, line 29: This is unfortunate. This public function is no longer used, but it's public, so we can't remove it.

dethi: You can mark it as depreacted and point to replacement in the function doc: https://go.dev/wiki/Deprecated

cluckett-arista: marked as deprecated.

PR #303, opened 2025-08-18
rpc: add individual operation observability tracking to SendBatch

vjadhav-arista:
Previously, SendBatch only tracked overall batch success/failure metrics. This change adds detailed observability for each individual operation in the batch, providing better visibility into operation-level performance and failure patterns.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -439,7 +439,8 @@ func (c *client) findClients(ctx context.Context, batch []hrpc.Call, res []hrpc.
 //     retryable RPCs may eventually succeed we need to return !ok to
 //     the caller of SendBatch.
 func (c *client) waitForCompletion(ctx context.Context, rc hrpc.RegionClient,
-	rpcs []hrpc.Call, results []hrpc.RPCResult, rpcToRes map[hrpc.Call]int, rpcStartTime time.Time) (
+	rpcs []hrpc.Call, results []hrpc.RPCResult, rpcToRes map[hrpc.Call]int,
+	rpcStartTime time.Time) (
 	retryables []hrpc.Call, shouldBackoff, unretryableError, ok bool) {
 
 	ok = true
--- a/rpc.go
+++ b/rpc.go
@@ -294,6 +294,7 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
 	table := batch[0].Table()
 	res = make([]hrpc.RPCResult, len(batch))
 	rpcToRes := make(map[hrpc.Call]int, len(batch))
+	rpcStartTime := time.Now() // Single start time for all RPCs in the batch
 	for i, rpc := range batch {
 		// map Call to index in res so that we can set the correct
 		// result as Calls complete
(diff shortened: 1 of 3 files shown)


rpc.go
```diff
@@ -459,8 +465,15 @@ loop:
 				case region.ServerError, region.NotServingRegionError:
 					retryables = append(retryables, rpc)
```
devoxel, line 466: Maybe we should track this as a retryable error. It would be good also to know if it was region.ServerError or NotServingRegionError.

My thinking is that from an SLI point of view, a request failure is a failure, it means we are erroring. If we let retriable errors escape, we miss monitoring.

The counterpoint to this is that our histogram gets filled with lots of labels and explodes cardinality.

What if we had a counter, not histogram, for requests that had a detailed label for error type, maybe in a follow up change? This would not track duration, just response type from HBase. I think that and this are probably separate, but both would add value.

rpc.go
```diff
@@ -475,14 +488,27 @@ loop:
...
+			} else {
+				// Track successful operation that completed after cancellation
+				o := operationDurationSeconds.WithLabelValues(description, "ok")
+				observability.ObserveWithTrace(ctx, o, duration)
 			}
 		default:
 			results[rpcToRes[rpc]].Error = ctx.Err()
+			// Track operation that was canceled
+			duration := time.Since(rpcStartTime).Seconds()
+			o := operationDurationSeconds.WithLabelValues(description, "error")
```
devoxel, line 510: In this case, I think "error" might be misleading to any SLI metrics; because in this case, we are not erroring because of HBase returning back a specific response, we are erroring because the context got cancelled.

devoxel (commented):
Generally approve, other than some nits.

I won't block on this, but I am holding back approval review because I think it needs a review from somebody who works more on gohbase

rpc.go
```diff
@@ -294,6 +294,7 @@ func (c *client) SendBatch(ctx context.Context, batch []hrpc.Call) (
 	table := batch[0].Table()
 	res = make([]hrpc.RPCResult, len(batch))
 	rpcToRes := make(map[hrpc.Call]int, len(batch))
+	rpcStartTime := time.Now() // Single start time for all RPCs in the batch
```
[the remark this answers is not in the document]
vjadhav-arista: okay fair enough.

rpc.go
vjadhav-arista: yes we have a follow up change https://github.com/tsuna/gohbase/pull/304, which is right candidate for HBase operations SLI 

current change is more suitable for insights into HBase operations from clients point of view - and also some estimation around how much time did an operation take including all the retries we had to do.

aaronbee (approved):

rpc.go
```diff
@@ -438,7 +439,7 @@ func (c *client) findClients(ctx context.Context, batch []hrpc.Call, res []hrpc.
 //     retryable RPCs may eventually succeed we need to return !ok to
 //     the caller of SendBatch.
 func (c *client) waitForCompletion(ctx context.Context, rc hrpc.RegionClient,
-	rpcs []hrpc.Call, results []hrpc.RPCResult, rpcToRes map[hrpc.Call]int) (
+	rpcs []hrpc.Call, results []hrpc.RPCResult, rpcToRes map[hrpc.Call]int, rpcStartTime time.Time) (
```
aaronbee, line 442: ./rpc.go:442: Line too long (101/100)

PR #304, opened 2025-08-19
region/client: Add per-operation RPC result metrics tracking

vjadhav-arista:
Implements detailed metrics tracking for RPC operations with support for both individual and batch (multi) RPCs. Tracks success/failure counts per operation type and region server.

### DIFF
--- a/region/client_test.go
+++ b/region/client_test.go
@@ -618,6 +618,7 @@ func TestReceiveDecodeProtobufError(t *testing.T) {
 	mockCall.EXPECT().ResultChan().Return(result).Times(1)
 	mockCall.EXPECT().NewResponse().Return(&pb.MutateResponse{}).Times(1)
 	mockCall.EXPECT().Context().Return(context.Background()).Times(1)
+	mockCall.EXPECT().Name().Return("Whatever").Times(1)
 
 	c.sent[1] = mockCall
 	c.inFlight = 1
@@ -666,6 +667,7 @@ func TestReceiveDeserializeCellblocksError(t *testing.T) {
 	mockCall.EXPECT().ResultChan().Return(result).Times(1)
 	mockCall.EXPECT().NewResponse().Return(&pb.MutateResponse{}).Times(1)
 	mockCall.EXPECT().Context().Return(context.Background()).Times(1)
+	mockCall.EXPECT().Name().Return("Get").Times(1)
 
 	c.sent[1] = callWithCellBlocksError{mockCall}
 	c.inFlight = 1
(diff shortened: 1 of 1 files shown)


region/client.go
```diff
@@ -517,7 +526,22 @@ func (c *client) receive(r io.Reader) (err error) {
 	// Here we know for sure that we got a response for rpc we asked.
 	// It's our responsibility to deliver the response or error to the
 	// caller as we unregistered the rpc.
-	defer func() { returnResult(rpc, response, err) }()
+	defer func() {
+		status := "success"
```
aaronbee, line 530: We use "ok" and "error" in other places.

vjadhav-arista: Ack. Corrected.

region/client.go
```diff
@@ -436,6 +436,15 @@ func (c *client) trySend(rpc hrpc.Call) (err error) {
 			c.fail(err)
 		}
 		if r := c.unregisterRPC(id); r != nil {
+			if m, ok := r.(*multi); ok {
```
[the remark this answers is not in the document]
vjadhav-arista: yeah makes sense.

region/client.go
```diff
@@ -517,7 +526,22 @@ func (c *client) receive(r io.Reader) (err error) {
 	// Here we know for sure that we got a response for rpc we asked.
 	// It's our responsibility to deliver the response or error to the
 	// caller as we unregistered the rpc.
-	defer func() { returnResult(rpc, response, err) }()
+	defer func() {
+		status := "success"
+		if err != nil {
+			status = "failure"
```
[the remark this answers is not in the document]
vjadhav-arista: Hmm yes that is a good idea, I will do that as a follow up change.

region/client.go
```diff
@@ -464,6 +465,20 @@ func (c *client) receiveRPCs() {
 	}
 }
 
+func (c *client) trackRPCResult(rpc hrpc.Call, status string) {
+	if m, ok := rpc.(*multi); ok {
+		for _, call := range m.calls {
+			if call != nil {
+				rpcResultCount.WithLabelValues(
+					c.Addr(), call.Name(), status, "batch").Inc()
```
aaronbee, line 473: For each call these labels are going to be the same all the time, call.Name() is going to be "Mutate" though technically could be a "Get".

I'd rather we do a single:
`rpcResultCount.WithLabelValues(c.Addr(), "Mutate", status, "batch").Add(len(m.calls))`

Because the WithLabelValues calls can be expensive.

To be more correct you could add up all the Mutate and the Get calls in a loop and then do the `Add` for each of them.

PR #310, opened 2025-10-22
Expose API's to manage replication peers through admin client

tomasbanet:
Add following new methods to admin client, each of which make an RPC to HMaster:
* ListReplicationPeers
* AddReplicationPeer
* RemoveReplicationPeer
* EnableReplicationPeer
* DisableReplicationPeer

for listing/creating/deleting & disabling/enabling HBase replication peers.

Updating peer config and other replication related RPC's are not implemented in this PR.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -3077,7 +3077,7 @@ func TestReplication(t *testing.T) {
 			removePeer := hrpc.NewRemoveReplicationPeer(ctx, peer.GetId())
 			err := ac.RemoveReplicationPeer(removePeer)
 			if err != nil {
-				t.Fatal(err)
+				t.Error(err)
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


integration_test.go
```diff
@@ -3039,3 +3039,196 @@ func consumeScanner(c gohbase.Client, scan *hrpc.Scan) ([]*hrpc.Result, error) {
...
+		}
+		peers, err := ac.ListReplicationPeers(listPeers)
+		if err != nil {
+			t.Fatal(err)
+		}
+		for _, peer := range peers {
+			removePeer := hrpc.NewRemoveReplicationPeer(ctx, peer.GetId())
+			err := ac.RemoveReplicationPeer(removePeer)
+			if err != nil {
+				t.Fatal(err)
```
grigorecv: should we not use instead Error and attempt to cleanup as most as possible?

hrpc/replication.go
```diff
@@ -0,0 +1,337 @@
...
+				"AddReplicationPeer with NewAddReplicationPeer() ?")
+		}
+		if table == nil || len(table) == 0 {
+			return errors.New("table must not be nil or empty")
+		}
+		l.tableCfs = append(l.tableCfs,
+			&pb.TableCF{
+				TableName: &pb.TableName{
+					// TODO: handle namespaces
+					Namespace: []byte("default"),
```
grigorecv: should we make this a global constant with the TODO and use it in the tests as well?

ciacono: This is used in some other places (ex. EnableTable, CreateTable, DeleteTable, DisableTable) - I think it would make sense to make it a global constant in those places too. But this could be done in a different PR.

integration_test.go
```diff
@@ -3039,3 +3039,196 @@ func consumeScanner(c gohbase.Client, scan *hrpc.Scan) ([]*hrpc.Result, error) {
...
+	}
+	for i, tc := range tcases {
+		testNoStr := strconv.Itoa(i)
+		t.Run("create list and invert enabled state of peer test #"+testNoStr, func(t *testing.T) {
+			peerId := peerIdPrefix + "_" + testNoStr
+			clusterKey := strings.Join([]string{
+				*host, "2181",
+				"/hbase-" + testNoStr,
+			}, ":")
+			tableBytes := []byte(table + "_" + testNoStr)
```
ciacono: Does the table not have to match the name of the actual table for this to work (re: adding the `testNoStr` suffix)?

integration_test.go
```diff
@@ -3039,3 +3039,196 @@ func consumeScanner(c gohbase.Client, scan *hrpc.Scan) ([]*hrpc.Result, error) {
...
+			}
+			peer := peers[0]
+			if peerId != peer.GetId() {
+				t.Fatalf("Expected peerId %q. Got: %q", peerId, peer.GetId())
+			}
+			if clusterKey != peer.GetConfig().GetClusterkey() {
+				t.Fatalf("Expected clusterKey %q. Got: %q", clusterKey,
+					peer.GetConfig().GetClusterkey())
+			}
+			if (tc.enabled && peer.GetState().GetState() != pb.ReplicationState_ENABLED) ||
```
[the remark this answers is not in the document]
ciacono: Agree on this, I think it makes it a lot easier to follow what's being validated. In this example, `peerState` could even be `peerReplicationState` to be more explicit.

tomasbanet:
PR was failing integration tests because after we add a replication peer in the test, the regionserver aborts some time after the TestReplication test passes, and the next test times out. 

2025-11-03 11:49:47,377 WARN  [RS_REFRESH_PEER-regionserver/localhost:16020-0.replicationSource,gohbase_test_1762170518739627000_0] client.ConnectionImplementation: Retrieve cluster id failed
2025-11-03 11:49:47,384 ERROR [RS_REFRESH_PEER-regionserver/localhost:16020-0.replicationSource,gohbase_test_1762170518739627000_0] regionserver.HRegionServer: ***** ABORTING region server 127.0.0.1,16020,1762170423636: Unexpected exception in RS_REFRESH_PEER-regionserver/localhost:16020-0.replicationSource,gohbase_test_1762170518739627000_0 *****
java.lang.IllegalStateException: Source should be active.
    at org.apache.hadoop.hbase.replication.regionserver.ReplicationSource.initialize(ReplicationSource.java:564)

Right now in TestReplication we're adding replication peers with fake sinks/destinations, as we don't have a secondary hbase cluster running as part of CI. So source will never be able to connect to the sink/destination RS, causing source RS to abort.

HBase 2.5.X fixes so that adding a peer for an invalid sink doesn't abort the regionserver - https://issues.apache.org/jira/browse/HBASE-27368

So I think we should merge the PR upgrading hbase to 2.6.2 to prevent this issue and then rebase the replication PR to run against hbase 2.6.2.

The alternative is running a second hbase cluster in CI with different ports and zkroot that is used  in TestReplication. But it seems overkill to run a second cluster just for testing a few replication methods on the admin client.

PR #322, opened 2025-11-21
Refactor scanner

aaronbee:
Move code around without making any behavior changes.

* Moved code from fetch() into request(), so that fetch's only job is retrying until getting a non-emptry result.

* Merged the code for isDone into update(). Now update has clear delineation between a scanner that needs to continue within the region, go to the next region, or is complete.

The goal was to reduce the need to jump between fetch(), update(), isDone() to understand how the scanner works.

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -55,21 +55,9 @@ type scannerV2 struct {
 func (s *scanner) fetch() (*hrpc.ScanResponseV2, error) {
 	// keep looping until we have error, some non-empty result or until close
 	for {
-		response, pbresp, region, err := s.request()
+		response, err := s.request()
 		if err != nil {
-			s.Close()
 			return nil, err
-		}
-		if s.rpc.TrackScanMetrics() && pbresp.ScanMetrics != nil {
-			metrics := pbresp.ScanMetrics.GetMetrics()
-			for _, m := range metrics {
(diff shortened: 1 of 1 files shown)


scanner.go
```diff
@@ -250,95 +238,136 @@ func (s *scannerV2) Next() (*hrpc.ScanResponseV2, error) {
...
+
+	return rpc.Response, nil
 }
 
-// update updates the scanner for the next scan request
+// update updates the scanner for the next scan request or closes it.
 func (s *scanner) update(resp *pb.ScanResponse, region hrpc.RegionInfo) {
-	if s.isRegionScannerClosed() && resp.ScannerId != nil {
-		s.openRegionScanner(resp.GetScannerId())
+	// Either there are more results in this region, more results in
```
devoxel: This is a behaviour change in the refactored version:

* we used to check for closure before checking for more results.
* I think Close could be called asynchronously by a client
* in that case, we would then recall openRegionScanner. 

There also is potentially some racyness with setting curRegionScannerID:
* if we close, it sets it to noScannerID,
* but if we are racing in this block, we could set it here instead

Finally, isDone also checks for closure, which is now folded into this block from what I understand

I don't know if either of these things are impactful, actually the first one sounds a bit more like a bug. I'm wondering if you thought about these cases.

PR #325, opened 2025-12-31
Scanner scan

aaronbee:
Add Scan method and return Region information in ScanResponse. This PR builds on top of #324

### DIFF
--- a/hrpc/scan.go
+++ b/hrpc/scan.go
@@ -75,6 +75,16 @@ type ScanResponseV2 struct {
 	// completed. It is the size of the uncompressed cellblocks in the
 	// response.
 	ResponseSize int
+
+	// Region is the region that was scanned for this response.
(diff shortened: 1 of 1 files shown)


scanner.go
```diff
@@ -229,19 +236,61 @@ func (s *scannerV2) Next() (*hrpc.ScanResponseV2, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return response, nil
+}
+
+// makeScanRequest creates the next scan request. If rows is non-zero,
+// then the number of rows to be read is overridden from the original
+// Scan.
```
stephane-arista: would it make sense to also mention that when rows is non zero, we set the option for how many rows we would like from hbase?

stephane-arista (approved):

ciacono (approved):

PR #328, opened 2026-01-20
Restore Renewer fixes. (#324) with startRow

ciacono:
startRow must be retained in the scanner renewal request because it is used for region lookup in SendRPC (even though the startRow is not used by HBase itself to renew the scanner).

### DIFF
--- a/scanner.go
+++ b/scanner.go
@@ -89,11 +89,12 @@ func (s *scanner) peek() (hrpc.ResultV2, error) {
 	if err != nil {
 		return hrpc.ResultV2{}, err
 	}
-	if !s.closed && s.rpc.RenewInterval() > 0 {
-		// Start up a renewer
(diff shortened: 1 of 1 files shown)


scanner.go
```diff
@@ -423,15 +425,14 @@ func (s *scanner) closeRegionScanner() {
...
+func (s *scanner) renew(ctx context.Context, scannerID uint64, startRow []byte) error {
 	if err := ctx.Err(); err != nil {
 		return err
 	}
 	rpc, err := hrpc.NewScanRange(ctx,
 		s.rpc.Table(),
-		startRow,
-		nil,
-		hrpc.ScannerID(s.curRegionScannerID),
+		startRow, nil,
```
dethi: Bit weird to have these two on the same line now

PR #329, opened 2026-01-22
CheckAndMutate

cluckett-arista:
Adds [`CheckAndMutate` to gohbase](https://hbase.apache.org/2.5/apidocs/org/apache/hadoop/hbase/client/CheckAndMutate.html). This will replace CheckAndPut in 3.x. More importantly, this allows even more functionality than the CheckAndPut (which allows comparison based on some comparator) by allowing some mutation other than a Put.

### DIFF
--- a/hrpc/hrpc_test.go
+++ b/hrpc/hrpc_test.go
@@ -1731,7 +1731,8 @@ func TestCheckAndMutate(t *testing.T) {
 			family:      "cf",
 			qualifier:   "q",
 			compareType: pb.CompareType_EQUAL,
-			comparator:  filter.NewBinaryComparator(filter.NewByteArrayComparable([]byte("expected"))),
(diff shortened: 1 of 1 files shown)


integration_test.go
```diff
@@ -1534,6 +1534,160 @@ func TestCheckAndPutParallel(t *testing.T) {
...
+	c := gohbase.NewClient(*host)
+	defer c.Close()
+
+	key := "row_cam_1"
+	cf := "cf"
+	qualifier := "a"
+
+	// First, put an initial value
+	initialValues := map[string]map[string][]byte{"cf": {"a": []byte("100")}}
+	putReq, err := hrpc.NewPutStr(context.Background(), table, key, initialValues)
```
ciacono: nit: can also use the helper method `insertKeyValueAtCol`

PR #333, opened 2026-02-26
Fixed TestChangingRegionServers to wait for region servers to come up

shrinidhi-arista:
TestChangingRegionServers wasn't waiting for regions to be created and moved, updated it to ensure the test waits for region servers to be changed.


ciacono (commented):
Thanks!

ciacono (approved):

PR #337, opened 2026-04-09
region: Add scan_queue_latency metric

aaronbee:
Add a histogram to track the amount of time spent waiting in the scan queue.

A heavily loaded region server may have concurrent scans limit drop to a minimal value, and we can observe this with existing metrics, but this only tells part of the story. To see the impact of the loaded regionserver we need the amount of time a request spends waiting to execute.

### DIFF
--- a/region/prometheus.go
+++ b/region/prometheus.go
@@ -70,6 +70,16 @@ var (
 		[]string{"regionserver"},
 	)
 
+	scanQueueLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "gohbase",
+			Name:      "scan_queue_latency_seconds",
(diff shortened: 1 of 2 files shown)


region/prometheus.go
```diff
@@ -70,6 +70,16 @@ var (
 		[]string{"regionserver"},
 	)
 
+	scanQueueLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "gohbase",
+			Name:      "scan_queue_latency",
```
alexaliu: We should include the unit in the metric name `scan_queue_latency_seconds`.

aaronbee: Done.

alexaliu (approved):

PR #344, opened 2026-05-27
Add QueueTime to ScanStats

aaronbee:
QueueTime is the time a Scan starts requesting a token from congestion control. Start.Sub(QueueTime) is the duration spent queued.


apchau825 (approved):
LGTM!

PR #345, opened 2026-06-03
Add Tests for CheckAndMutate ifNotExists conditions

cluckett-arista:
Adds tests for multi-column CheckAndMutate calls with nil comparators. Whereas CheckAndPut with a `nil` comparator value is treated as "write if not exists", CheckAndMutate does not behave the same way. Regardless of filterIfMissing, if none of the keys exist, then the values are not written. If at least one of the keys exist, then the values will be written (provided filterIfMissing=false). This is an HBase implementation. Instead, one has to create a ifNotExists filter in order for these changes to be applied, which represent the other half of the tests introduced in this commit.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -1705,6 +1705,103 @@ func TestCheckAndMutateDelete(t *testing.T) {
 	}
 }
 
+func TestCheckAndMutateIfNotExistsMultipleColumns(t *testing.T) {
+	c := gohbase.NewClient(*host)
+	defer c.Close()
+
+	cf := "cf"
+	mutationQual := "target"
+	newData := []byte("newData")
+	checkColumns := []string{"key1", "key2"}
+
+	checkFilters := make([]filter.Filter, len(checkColumns))
+	for i, col := range checkColumns {
+		checkFilters[i] = columnNotExistsFilter(cf, col)
+	}
+	condFilter := filter.NewList(filter.MustPassAll, checkFilters...)
+
+	tests := []struct {
+		name          string
+		setupData     map[string][]byte
+		shouldProcess bool
+		storedVal     []byte
+	}{
+		{
+			name:          "none of the keys exist yet, will store",
+			setupData:     map[string][]byte{"other": []byte("someNewValue")},
+			shouldProcess: true,
+			storedVal:     newData,
+		},
+		{
+			name:          "one of the checked keys already exists, will not store",
+			setupData:     map[string][]byte{"key1": []byte("aValueToSomeExistingKey")},
+			shouldProcess: false,
+		},
+		{
+			name: "both of the checked keys already exists, will not store",
+			setupData: map[string][]byte{
+				"key1": []byte("aValueToSomeExistingKey"),
+				"key2": []byte("bValueToSomeExistingKey"),
+			},
+			shouldProcess: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			key := fmt.Sprintf("row_cam_ifnotexists_multi_%d", time.Now().UnixNano())
+
+			putReq, err := hrpc.NewPutStr(context.Background(), table, key,
+				map[string]map[string][]byte{cf: tt.setupData})
+			if err != nil {
+				t.Fatal(err)
+			}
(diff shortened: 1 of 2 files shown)


ciacono:
> Adds tests for multi-column CheckAndMutate calls with nil comparators. Whereas CheckAndPut with a `nil` comparator value is treated as "write if not exists", CheckAndMutate does not behave the same way. Regardless of filterIfMissing, if none of the keys exist, then the values are not written. If at least one of the keys exist, then the values will be written (provided filterIfMissing=false). This is an HBase implementation. Instead, one has to create a ifNotExists filter in order for these changes to be applied, which represent the other half of the tests introduced in this commit.

So this change is to add a ifNotExists filter so that CheckAndMutate can have the same behaviour as CheckAndPut  where it can write if not exists if there is a nil comparator, correct? Since HBase CheckAndMutate doesn't offer this while CheckAndPut does?

integration_test.go
```diff
@@ -1822,20 +1921,73 @@ func TestCheckAndMutateWithFilter(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("%s: %v", tt.name, err)
+		rowKey := key
+		if tt.rowKey != "" {
+			rowKey = tt.rowKey
 		}
-		if _, err = c.Put(putReq); err != nil {
-			t.Fatalf("%s: %v", tt.name, err)
+
+		if !tt.skipSetup {
```
ciacono: Nit: it could be a bit more readable to check if len(setupData) == 0, then don't need this extra variable to check or set in the tcase.

integration_test.go
```diff
@@ -1822,20 +1921,73 @@ func TestCheckAndMutateWithFilter(t *testing.T) {
...
+					false, false),
+			),
+			mutationVal:    []byte("never_stored"),
+			wantProcessed:  false,
+			wantFinalValue: nil,
+		},
 	}
 
 	// Note that these cases were meant to be run sequentially, not in parallel
 	for _, tt := range tests {
```
ciacono, line 1971: Could use t.Run here so t.Fatalf wouldn't stop other testcases from being run

integration_test.go
```diff
@@ -1822,20 +1921,73 @@ func TestCheckAndMutateWithFilter(t *testing.T) {
...
+				filter.NewSingleColumnValueFilter(
+					[]byte(cf), []byte("key2"),
+					filter.Equal,
+					filter.NewBinaryComparator(
+						filter.NewByteArrayComparable([]byte("v2"))),
+					false, false),
+			),
+			mutationVal:    []byte("surprise_stored"),
+			wantProcessed:  true,
+			wantFinalValue: []byte("surprise_stored"),
```
ciacono: I think it would be a bit more clear about this test case for future readers adding a comment why this is a surprise_stored

integration_test.go
```diff
@@ -2018,6 +2170,15 @@ func insertKeyValue(c gohbase.Client, key, columnFamily string, value []byte,
 	return insertKeyValueAtCol(c, key, "a", columnFamily, value, options...)
 }
 
+// columnNotExistsFilter returns a filter that passes when the given column is absent.
+func columnNotExistsFilter(columnFamily, qualifier string) filter.Filter {
```
ciacono, line 2174: Note that this would pass also if the column is present but the qualifier is empty bytes, I think this is a HBase side limitation so should document.

ciacono (commented):
LGTM, a few small comments

integration_test.go
cluckett-arista: that was a terrible name for a value, renamed

integration_test.go
cluckett-arista: agreed

cluckett-arista:
> So this change is to add a ifNotExists filter so that CheckAndMutate can have the same behaviour as CheckAndPut where it can write if not exists if there is a nil comparator, correct? Since HBase CheckAndMutate doesn't offer this while CheckAndPut does?

One would expect CheckAndMutate to act like CheckAndMutate, and it *mostly* does. What caught us off guard is a specific edge case. It will write with a nil cmp when at least one key exists (expected), but will *not* write when no key exists yet (unexpected). That's the HBase functionality that we're shoehorned into.

ciacono (approved):

PR #346, opened 2026-06-08
Add NameBytesPair to RequestHeader

cluckett-arista:
[NameBytesPair exists in RequestHeader in the HBase RPC](
https://github.com/apache/hbase/blob/ab36c816391a211f68c928f124a9e8697f6a4a58/hbase-protocol-shaded/src/main/protobuf/rpc/RPC.proto#L152), but we were not exposing it. This adds it.

There are a few small things to note.
1. `attribute` in the HBase proto is a repeated field, so we use `attributes` instead to represent plurality.
2. There exist `attribute` in several places. In the event we add the `attribute` from `ConnectionHeader`, we will have a naming conflict, so we use `HeaderAttributes` instead to allow distinguishing them later.
3. HBase is the consumer, so gohbase doesn't offer the applicable unmarshal functions. In order for codecov to be satisfied, we need to add tests. The only test that really makes sense is a roundtrip proto test, so we're forced to define a quick unmarshal/decode function to assist.


cluckett-arista:
sorry, missed the marshalProto call. Will update that.

PR #347, opened 2026-06-08
Add NameBytesPair to RequestHeader

cluckett-arista:
[NameBytesPair exists in RequestHeader in the HBase RPC](https://github.com/apache/hbase/blob/ab36c816391a211f68c928f124a9e8697f6a4a58/hbase-protocol-shaded/src/main/protobuf/rpc/RPC.proto#L152), but we were not exposing it. This adds it.

There are a few small things to note.
* `attribute` in the HBase proto is a repeated field, so we use `attributes` instead to represent plurality.
* There exist `attribute` in several places. In the event we add the attribute from `ConnectionHeader`, we will have a naming conflict, so we use `HeaderAttributes` instead to allow distinguishing them later.
* HBase is the consumer, so gohbase doesn't offer the applicable unmarshal functions. In order for codecov to be satisfied, we need to add tests. The only test that really makes sense is a roundtrip proto test, so we're forced to define a quick unmarshal/decode function to assist.


region/client_test.go
```diff
@@ -1454,6 +1474,83 @@ func TestScanConcurrencyControlSetCapacity(t *testing.T) {
 	}
 }
 
+func TestMarshalProtoHeaderAttributes(t *testing.T) {
+	// TODO: Expand to add broader request header fields
```
ciacono: What would be some examples of the broader request header fields? What would be some actual attributes?

hrpc/query.go
```diff
@@ -115,6 +116,40 @@ func GetPriority(c Call) uint32 {
...
+
+// HeaderAttribute is an option that sets a key-value attribute on the RPC request header.
+// It can be used with Get and Scan requests. HeaderAttribute can be specified multiple times
+// and attributes will be accumulated.
+func HeaderAttribute(key string, val []byte) func(Call) error {
+	return func(c Call) error {
+		bq, ok := c.(interface {
+			setHeaderAttribute(key string, val []byte)
+		})
+		if !ok {
```
ciacono, line 141: Could add tests to validate this is only for Get and Scan reqs?

ciacono (commented):
LGTM

region/client_test.go
cluckett-arista: TBH, this is going to be implementation-specific. this comment should be removed at this point since I added it when I only had one test case. I didn't remove it when I came back and added more.

hrpc/query.go
cluckett-arista: uh... there's a misunderstanding here. There exist Get/Scan/Mutate requests that , but AFAIK this kind of `attribute` belongs to the request header, which can be passed for *any* call. The naming in the HBase API is admittedly confusing to follow because there are several `attribute`s.

cluckett-arista: well apparently this does not get added to Put requests. That wasn't super clear to me from reading the documentation. So it acts like `Priority`.

ciacono (approved):

